PR #27, opened 2015-01-13
Include the workertype in list of pending tasks

jhford:
For the provisioner, we need to know not the number of pending tasks but rather
the number of pending tasks for each worker type.  As the api is implemented
now, we need to fetch the list of all pending tasks for the provisioner in
question, then do a /task/:taskId for each taskId which is pending to determine
which workerType a given task needs.

  Let's instead includ this information in the return value from the get
  pending tasks endpoint.  The cost is JSON overhead and string data with a
  benefit of reducing the number of API calls substantially.

### DIFF
--- a/queue/task.js
+++ b/queue/task.js
@@ -443,7 +443,10 @@ Task.queryPending = transacting(function(provisionerId, knex) {
       'runs.state':           'pending'
     }).then(function(tasks) {
       return tasks.map(function(task) {
-        return slugid.encode(task.taskId);
+        return {
+          taskId: slugid.encode(task.taskId),
+          workerType: task.workerType
+        }
       });
     });
 });


jonasfj:
This seems to break: https://github.com/taskcluster/taskcluster-queue/blob/769d07ffdd7028d74654d770ad2bb52c7d40f48b/routes/api/v1.js#L1198-L1202

Either way, perhaps we can agree to return something like this instead:
```js
{
   "MyWorkerType":    24,
   "AnotherWorkerType": 50
}
```
As a mapping from workerTypes to number of pending tasks. We plan to migrate the queue away from postgres to azure table storage + azure queue storage... Hence, we'll store pending tasks in an azure queue, that means we can get an estimate of pending tasks (a pretty good estimate)... But we can't iterate through all pending tasks (or at least we won't want to).

Note, I think the list of pending tasks already did specify workerType, the list is basically a list of task status structures. But it would be better if we moved to the simple reply form where we don't list tasks... but just give the numbers for the workerTypes, agree?

jhford:
I support #28 instead of this, closing this PR

PR #28, opened 2015-01-14
Added end-point to get summary of pending task, for various workerTypes

jonasfj:
@jhford I hope this does what you wanted to achieve from PR #27.
I think this end-point provides the information needed by the aws-provisioner, and I'm pretty sure we can reimplement this end-point once we start storing the queue/list of pending tasks in azure queue storage.

I didn't add any schema, but the return value is:
```js
{
  "<workerType>": <number-of-pending-tasks>,
  ....
}
```

For all workerTypes with pending tasks, for the given `provisionerId`. Note workerTypes that don't have pending tasks will not be mentioned...

**Remark**, I have a small fear that when moving the azure queue storage, we'll have to include the list of workerTypes you want to inquire about in the request. Though this should not be a problem the provisioner to do... arrhh... As of writing I realize that we'll probably have to do it that way, once we move azure queue storage... Ahh, **** I'll just add that end-point too while I'm at it... :)


jonasfj:
Pushed new code with the end-point `pendingTask` and `pendingTaskCount`.

I think `pendingTask` is future proof. Is this crazy :)

@jhford,
When you okay this we can remove `pendingTaskCount` and merge it...
We need to keep `getPendingTasks` until we've deployed the aws-provisioner you're working on.


Unrelated remark, I'll start work on porting the queue to azure table/queue storage sometime, not too far from now...

jhford:
I like this endpoint much better!  This looks merge-able to me, but as I'm not too familar with the Queue, I'll let you push the merge button.

For my purposes, having 100% accuracy is not a requirement.  I think if we switch to a data source that does not make this guarantee, we should probably throw that in the documentation now.

jonasfj:
Wow, I never actually expected travis to work... that's awesome... merged...

PR #29, opened 2015-01-23
Migration to azure queue

jonasfj:
New task polling flow.... Not as well-documented as it could be... but the flow for workers is covered in details here:
http://docs.taskcluster.net/queue/worker-interaction/

Most notable changes are in `v1.js` specifically `claimTask` and `taskPollUrls`.

**Warning**, this change carries configuration changes, **please NOT merge**. Just give an `r+`.


jonasfj:
Note to self for deployment of this, we must specify the environment variable `queue_signatureSecret` as some secret only known the server...

PR #30, opened 2015-01-28
Bug 1080286 - Added cancel endpoint

gregarndt:


routes/api/v1.js
```diff
@@ -1079,6 +1079,88 @@ api.declare({
...
+/** Report task canceled */
+api.declare({
+  method:     'post',
+  route:      '/task/:taskId/runs/:runId/canceled',
+  name:       'reportCanceled',
+  scopes: [
+    [
+      'queue:resolve-task',
+      'assume:worker-id:<workerGroup>/<workerId>'
+    ]
```
jonasfj: I see some copy/paste here :)

Let's set the scope set to:
 * `queue:cancel-task`
 * `assume:scheduler-id:<schedulerId>/<taskGroupId>`
Naturally this also changes the parameters to the `req.satisfies` call.

The scope `assume:scheduler-id:<schedulerId>/<taskGroupId>` is also required to rerun a task. So I think this is an appropriate granularity. After all this end-point is supposed to be used by _schedulers_ that want's to cancel a task they have scheduled.

routes/api/v1.js
```diff
@@ -1079,6 +1079,88 @@ api.declare({
 });
 
 
+/** Report task canceled */
+api.declare({
+  method:     'post',
+  route:      '/task/:taskId/runs/:runId/canceled',
+  name:       'reportCanceled',
```
jonasfj: Hmm... I'm not sure we should require a `runId` to be specified. After all don't we just want to cancel the latest run, regardless of it's `runId`.... Calling `queue.status` to get the `runId` and then calling `queue.reportCanceled` opens the possibility that a new run is created in between the two calls.. Note, the semantics are different based on whether or not we specify `runId`, so the argument for can definitely be made for both.

But I think the semantics we're looking for here is that a scheduler (task-graph scheduler) or person who created a task want's to cancel it, regardless of the current `runId`. Hence, I suggest route and name as follows:
```
  route:  '/task/:taskid/cancel',
  name: 'cancelTask'
```

routes/api/v1.js
```diff
@@ -1079,6 +1079,88 @@ api.declare({
...
+    // Authenticate request by providing parameters
+    if(!req.satisfies({
+      workerGroup:  workerGroup,
+      workerId:     workerId
+    })) {
+      return;
+    }
+
+    // Resolve task run
+    return ctx.Task.resolveTask(
```
jonasfj: Unfortunately it seems `Task.resolveTask` only works if the run is `running`. Which might not be the case.

PR #32, opened 2015-02-13
Rewrite on azure storage services

jonasfj:
This is a pretty big rewrite... I don't recommend trying to read the diff... Just look at the result..

**Please**, don't hit the merge button... We shouldn't land this before we've ported both provisioning logic and docker-worker. I'll backport non-deprecated APIs so that we can do this before we deploy this.

Anyways, please leave comments if you have questions... Or if anything looks not right, sketchy just ask...

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -244,7 +244,7 @@
               "dependencies": {
                 "ansi-regex": {
                   "version": "1.1.1",
-                  "from": "ansi-regex@>=1.0.0 <2.0.0",
+                  "from": "ansi-regex@>=1.1.0 <2.0.0",
                   "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-1.1.1.tgz"
                 },
                 "get-stdin": {
@@ -261,7 +261,7 @@
               "dependencies": {
                 "ansi-regex": {
                   "version": "1.1.1",
-                  "from": "ansi-regex@>=1.0.0 <2.0.0",
+                  "from": "ansi-regex@>=1.1.0 <2.0.0",
                   "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-1.1.1.tgz"
                 }
               }
@@ -1154,8 +1154,9 @@
       }
     },
     "taskcluster-base": {
-      "version": "0.7.8",
-      "from": "taskcluster-base@0.7.8",
+      "version": "0.7.9",
+      "from": "taskcluster-base@0.7.9",
+      "resolved": "https://registry.npmjs.org/taskcluster-base/-/taskcluster-base-0.7.9.tgz",
       "dependencies": {
         "express": {
           "version": "4.9.0",
@@ -2162,7 +2163,7 @@
               "dependencies": {
                 "mime-types": {
                   "version": "2.0.10",
-                  "from": "mime-types@>=2.0.4 <2.1.0",
+                  "from": "mime-types@>=2.0.9 <2.1.0",
                   "resolved": "https://registry.npmjs.org/mime-types/-/mime-types-2.0.10.tgz",
                   "dependencies": {
                     "mime-db": {
@@ -3109,7 +3110,7 @@
                 },
                 "inherits": {
                   "version": "2.0.1",
-                  "from": "inherits@>=2.0.1 <2.1.0",
+                  "from": "inherits@>=2.0.0 <3.0.0",
                   "resolved": "https://registry.npmjs.org/inherits/-/inherits-2.0.1.tgz"
                 }
               }
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "lodash":               "^3.0.0",
     "promise":              "6.1.0",
     "slugid":               "1.0.3",
-    "taskcluster-base":     "0.7.8",
+    "taskcluster-base":     "0.7.9",
     "thirty-two":           "^0.0.2",
     "url-join":             "0.0.1",
     "uuid":                 "2.0.1",
--- a/test/api/deadline_test.js
+++ b/test/api/deadline_test.js
@@ -16,7 +16,7 @@ suite('Deadline expiration (deadline-reaper)', function() {
       workerType:       'test-worker',
                         // Legal because we allow a small bit of clock drift
       created:          taskcluster.fromNowJSON('- 5 seconds'),
-      deadline:         taskcluster.fromNowJSON('5 second'),
+      deadline:         taskcluster.fromNowJSON('10 seconds'),
       retries:          1,
       payload:          {},
       metadata: {


jonasfj:
Oh, and don't worry about this failing on travis... I think it's due to network issues... sometimes it takes 4 mins on travis, sometimes it takes 30 min and times out left and right...

Note locally it only takes 90s to run `npm test`.

routes/v1.js
```diff
@@ -0,0 +1,1567 @@
...
+      newRun.state          === 'pending' &&
+      newRun.reasonCreated  === 'retry') {
+    await Promise.all([
+      this.queueService.putPendingMessage(task, runId),
+      this.publisher.taskPending({
+        status:         status,
+        runId:          runId + 1
+      }, task.routes)
+    ]);
+  } else {
```
gregarndt: Would we not want to publish to task-exception for a particular run that was caused by a worker shutdown?

routes/v1.js
```diff
@@ -0,0 +1,1567 @@
...
+  method:     'get',
+  route:      '/task/:taskId',
+  name:       'task',
+  idempotent: true,
+  scopes:     undefined,
+  output:     SCHEMA_PREFIX_CONST + 'task.json#',
+  title:      "Get Task Definition",
+  description: [
+    "This end-point will return the task-definition. Note, that the task",
+    "definition may have been modified by "
```
gregarndt: modified by.....

routes/v1.js
jonasfj: Not if we schedule another run to `retry`. At the moment we don't publish `task-exception` if the claim expired and we still have retries left, then we schedule a new run with `reasonCreated: 'retry'`. I think the same logic applies here, ie. we do what we normally would do when a claim is expired.

Note, IMO, `task-completed`, `task-exception` and `task-failed` should only come once for each task. Specifically when the queue is done processing the task. No more retries, etc. If some day we move auto-reruns into the queue it should do the same.

If we need run-level events, let's add them later as `run-exception`, etc...

gregarndt: Makes sense, thanks for clearing that up.  Right now in a few different places, we rely on those pulse message to indicate a run of a task is complete, rather than information about the task itself.  Having a way to listen for when a run has been completed could be useful later, but at least I know this is the behavior for now.  Thanks.

routes/v1.js
jonasfj: Nice catch, I'm on it :)

routes/v1.js
jonasfj: Well, a run being `completed` also results in the task being resolved as `completed`.. We don't schedule retries for successful task yet -- I'm sure someone will ask for this feature later, he he :)

Procfile
```diff
@@ -1,4 +1,5 @@
-web:              node bin/server.js            production
-reaper:           node bin/reaper.js            production
-expireArtifacts:  node bin/expire-artifacts.js  production
-retireTasks:      node bin/retire-tasks.js      production
\ No newline at end of file
+web:              npm run start
+claimReaper:      npm run claim-reaper
+deadlineReaper:   npm run deadline-reaper
+expireArtifacts:  npm run expire-artifacts
+expireTask:       npm run expire-tasks
```
jhford: Are all of these reapers/expirers such high load that they need their own complete instance?

README.md
```diff
@@ -64,6 +64,18 @@ With the following processes running as cron jobs on daily basis:
 
 On heroku these are configured using the scheduler.
 
+Monitoring
+----------
+This server will print log messages it is recommend run with `DEBUG` as
+`"* -superagent -6to5 -mocha:* -express:*"`.
+Notice that messages printed containing the substring `[alert-operator]` should
```
jhford: I like having `alert-operator`, but it's really critical that we don't use it as the basis for how we monitor and track the systems.  I know it's hard to do anything better without writing a real logging system, but I'm just a little nervous about establishing a convention of a string appearing in logs as being something that's a critical error.  At the least, could we avoid typos by exporting this string from taskcluster-base?  I know that sounds silly, but it'd suck if someone printed the string `'[alert-opreator]'` and we missed a critical issue because of it.

bin/claim-reaper.js
```diff
@@ -0,0 +1,129 @@
...
+        exchangePrefix:     cfg.get('queue:exchangePrefix'),
+        validator:          validator,
+        referencePrefix:    'queue/v1/exchanges.json',
+        drain:              influx,
+        component:          cfg.get('queue:statsComponent'),
+        process:            'claim-reaper'
+      });
+    })(),
+    Task.ensureTable()
+  ]);
```
jhford: Purdy es6 :)

queue/bucket.js
```diff
@@ -63,6 +63,14 @@ Bucket.prototype.createPutUrl = function(prefix, options) {
 };
 
 /**
+ * Create an unsigned GET URL
+ */
+Bucket.prototype.createGetUrl = function(prefix) {
+  assert(prefix,                "prefix must be given");
+  return `${this.s3.endpoint.href}${this.bucket}/${prefix}`;
```
jhford: is this for real?  Google searches suggest yes.  This makes me happy to see in JS.

queue/claimresolver.js
```diff
@@ -0,0 +1,243 @@
...
+    }
+    this.stopping = false;
+
+    // Start a loop for the amount of parallelism desired
+    var loops = [];
+    for(var i = 0; i < this.parallelism; i++) {
+      loops.push(this.poll());
+    }
+    // Create promise that we're done looping
+    this.done = Promise.all(loops).catch((err) => {
```
jhford: This means that if we have 4 loops running and one of them stalls, that all processing stops until the stalled loop ends.

Maybe we could have something in each loop that sets a timeout to ensure that an individual loop will throw an error if nothing happens in it for a reasonably long amount of time.

queue/claimresolver.js
```diff
@@ -0,0 +1,243 @@
...
+
+  /** Poll for messages and handle them in a loop */
+  async poll() {
+    while(!this.stopping) {
+      var messages = await this.queueService.pollClaimQueue();
+      debug("Fetched %s messages", messages.length);
+
+      await Promise.all(messages.map((message) => {
+        // Don't let a single task error break the loop, it'll be retried later
+        // as we don't remove message unless they are handled
```
jhford: do we have facilities for tracking which tasks have failed, disappeared from the API and are never seen again?  How do those things get resolved?

queue/claimresolver.js
```diff
@@ -0,0 +1,243 @@
...
+      run.state           = 'exception';
+      run.reasonResolved  = 'claim-expired';
+      run.resolved        = new Date().toJSON();
+
+      // Do **NOT** clear takenUntil on task, as this will prevent the message
+      // from running again. In case something below fails.
+
+      // If the run isn't the last run, then something is very wrong
+      if (task.runs.length - 1 !== runId) {
+        debug("[alert-operator] running runId: %s, resolved exception, " +
```
jhford: This sounds like a pretty major failure...  Do we really want to just log and ignore?

queue/deadlineresolver.js
```diff
@@ -0,0 +1,201 @@
...
+    }
+    this.stopping = false;
+
+    // Start a loop for the amount of parallelism desired
+    var loops = [];
+    for(var i = 0; i < this.parallelism; i++) {
+      loops.push(this.poll());
+    }
+    // Create promise that we're done looping
+    this.done = Promise.all(loops).catch((err) => {
```
jhford: Like above, should we have some sort of fail safe that ensures that no individual loop gets jammed up?  It sounds like these messages are only available for a defined amount of time, so we should probably be careful not to let the queue spin for too long if one of these loops gets stuck

queue/queueservice.js
```diff
@@ -5,63 +5,280 @@ var debug       = require('debug')('queue:queue');
...
+    // Ensure the claim queue exists
+    await this.ensureClaimQueue();
+
+    // Get messages
+    var messages = await new Promise((accept, reject) => {
+      this.service.getMessages(this.claimQueue, {
+        numOfMessages:        32,
+        peekOnly:             false,
+        visibilityTimeout:    10 * 60,
+      }, (err, messages) => { err ? reject(err) : accept(messages) });
```
jhford: This is really neat syntactical sugar, but it's a little hard to read at first.

queue/queueservice.js
```diff
@@ -5,63 +5,280 @@ var debug       = require('debug')('queue:queue');
...
+        }
+      };
+    });
+  }
 
-  // Find promise
-  var retval = this.queues[id];
+  /** Ensure existence of a queue */
+  ensurePendingQueue(provisionerId, workerType) {
+    // Construct id, note that slash cannot be used in provisionerId, workerType
```
jhford: `+ ensurePendingQueue(provisionerId, workerType) {'

is this valid syntax?  I'm not familiar with it if it is...

routes/artifacts.js
```diff
@@ -0,0 +1,597 @@
...
+    "automatically take care of deleting artifacts past their",
+    "expiration point. This features makes it feasible to upload large",
+    "intermediate artifacts from data processing applications, as the",
+    "artifacts can be set to expire a few days later.",
+    "",
+    "We currently support 4 different `storageType`s, each storage type have",
+    "slightly different features and in some cases difference semantics.",
+    "",
+    "**S3 artifacts**, is useful for static files which will be stored on S3.",
+    "When creating an S3 artifact is create the queue will return a pre-signed",
```
jhford: This sentence doesn't really parse...  "When creating an S3 artifact the queue will return..."?

routes/artifacts.js
```diff
@@ -0,0 +1,597 @@
...
+    "live log. A request to create an Azure artifact will return a URL",
+    "featuring a [Shared-Access-Signature]" +
+    "(http://msdn.microsoft.com/en-us/library/azure/dn140256.aspx),",
+    "refer to MSDN for further information on how to use these.",
+    "",
+    "**Reference artifacts**, only consists of meta-data which the queue will",
+    "store for you. These artifacts really only have a `url` property and",
+    "when the artifact is requested the client will be redirect the URL",
+    "provided with a `303` (See Other) redirect. Please note that we cannot",
+    "delete artifacts you upload to other service, we can only delete the",
```
jhford: Maybe we should include this warning in the description of the other services as well.  Duplication, but if someone's only looking for azure info, they'll likely only look at the azure section

routes/artifacts.js
```diff
@@ -0,0 +1,597 @@
...
+        status:   task.status(),
+        runId:    runId
+      }
+    });
+  }
+
+  // Construct details for different storage types
+  var isPublic = /^public\//.test(name);
+  var details  = {};
+  if (storageType === 's3') {
```
jhford: why not a switch with a default handler that return an error?

routes/artifacts.js
```diff
@@ -0,0 +1,597 @@
...
+    return res.reply({
+      storageType:  's3',
+      contentType:  artifact.contentType,
+      expires:      expiry.toJSON(),
+      putUrl:       putUrl
+    });
+  }
+
+  // Reply with SAS for azure
+  if (artifact.storageType === 'azure') {
```
jhford: Shouldn't this be `else if` or a `switch`?

routes/artifacts.js
```diff
@@ -0,0 +1,597 @@
...
+    }));
+  }
+
+  // Handle redirect artifacts
+  if (artifact.storageType === 'reference') {
+    return res.redirect(303, artifact.details.url);
+  }
+
+  // Handle error artifacts
+  if (artifact.storageType === 'error') {
```
jhford: I wonder if there's value to creating a set of objects, S3, Azure, Reference and Error which each know how to do each operation and set a variable once in the beginning of the function that says which object to call.  It seems like a great place for bugs to slip in where we have these repeating branches

test/api/artifact_test.js
```diff
@@ -11,23 +11,21 @@ suite('Post artifacts', function() {
...
   var data          = require('../../queue/data');
-  var helper        = require('./helper')();
-
-  // Create datetime for created and deadline as 3 days later
-  var created = new Date();
-  var deadline = new Date();
-  deadline.setDate(created.getDate() + 3);
+  var base          = require('taskcluster-base');
+  var taskcluster   = require('taskcluster-client');
+  var expect        = require('expect.js');
```
jhford: should.js is similar to this and reads much nicer, I think.

Procfile
jonasfj: yeah, it's easiest to scale that way...

Note, we might make expireArtifacts a part of expireTask in the future...

README.md
jonasfj: Yeah, we should do something in taskcluster-base that wraps this with `console.error("[alert-operator]" + msg, args)`.
This way we can avoid typos and we can't hide them with `DEBUG` env var...

Filed bug 1144521 to discuss/implement a proper method for this...

queue/bucket.js
jonasfj: Yes this is for real :)

Start using it in the provisioner :)

queue/claimresolver.js
jonasfj: A better option would be have to producer/consumer pattern... where the producer pulls from azure queue and consumer handles x number of messages in parallel.

Because the issue here is that we want to run x handle of messages in parallel... This is quick hack until I figure out a good pattern for how to do this nicely...

queue/claimresolver.js
jonasfj: If the nessage fails here it won't be removed from the azure queue... so it'll show up again after a time out... and then [alert-operator] should alert me...

queue/claimresolver.js
jonasfj: what else can we do... Crash isn't better...

It's an invariant violation. I seriously doubt it's going to happen... If it's happens it's probably a freak task that we want to investigate. If happens all the time we're well not good no matter what we do :)

jonasfj: invariant violation -> undefined behaviour

queue/deadlineresolver.js
jonasfj: It only does azure operations... They make time out... but as all of them are in paralllel ,I suspect we'll be fine.. But yeah, we should do a producer/consumer pattern...

And we should probably enforce timeout for azure table operations at a higher level, like in `base.Entity` that way we can maintain defined behaviour in cases like that.

queue/queueservice.js
jonasfj: Yeah, don't look at the diff... changes are too big :)

routes/artifacts.js
jonasfj: The warning says: If you upload a link as an artifact, we cannot delete the resource the link points to, pretty obvious :)

routes/artifacts.js
jonasfj: Switch statement didn't get shorter... and I generally try to keep indentation limited..
But in this case I seems to make sense...

jonasfj: Okay, I couldn't help not trying this.. But one could also do:
```js
// with switch statement 
switch(storageType) {
    case 's3':
      if (isPublic) {
        details.bucket  = this.publicBucket.bucket;
      } else {
        details.bucket  = this.privateBucket.bucket;
      }
      details.prefix    = [taskId, runId, name].join('/');
      break;

    case 'azure':
      details.container = this.blobStore.container;
      details.path      = [taskId, runId, name].join('/');
      break;

    case 'reference':
      details.url       = input.url;
      break;

    case 'error':
      details.message   = input.message;
      details.reason    = input.reason;
      break;

    default:
      throw new Error("Unknown storageType: " + storageType);
  }


/// With es6 dictionary:
{
    s3() {
      if (isPublic) {
        details.bucket  = this.publicBucket.bucket;
      } else {
        details.bucket  = this.privateBucket.bucket;
      }
      details.prefix    = [taskId, runId, name].join('/');
    }
    azure() {
      details.container = this.blobStore.container;
      details.path      = [taskId, runId, name].join('/');
    }
    reference() {
      details.url       = input.url;
    }
    error() {
      details.message   = input.message;
      details.reason    = input.reason;
    }
  }[storageType]();


// If writing a custom utility function called Switch implemented as
var Switch = (choice, options) => { return options[choice](); };
Switch(storageType, {
    s3() {
      if (isPublic) {
        details.bucket  = this.publicBucket.bucket;
      } else {
        details.bucket  = this.privateBucket.bucket;
      }
      details.prefix    = [taskId, runId, name].join('/');
    }
    azure() {
      details.container = this.blobStore.container;
      details.path      = [taskId, runId, name].join('/');
    }
    reference() {
      details.url       = input.url;
    }
    error() {
      details.message   = input.message;
      details.reason    = input.reason;
    }
  });
```

I love ES6 :)
Granted I went with switch statement, it's the decent thing to do here :)

routes/artifacts.js
jonasfj: Moved to switch but really sure it's any better...

I like the return early pattern, and use else and elseif unless I strictly have to...

routes/artifacts.js
jonasfj: It's tempting... but we currently don't support this on Entities... I'm not sure how we would...

If we want up with many artifact types... we should implement this as a strategy pattern... Where we have dictionary mapping from artifact type to artifact strategy that implements all the functions... Then this could be done at entity level or something... But there so many variables flying around here I'm not sure it's easy...

test/api/artifact_test.js
jonasfj: `should.js` is nice... but `should(a).be.equal.to(b)` reads wrong. And I don't like `a.should.be.equal.to(b)` because it modifies prototype.

What do you think of `assume` https://www.npmjs.com/package/assume
I'm starting to consider it for new stuff... I think it's a good name, more updated than `expect`...

But also reads nicely without polluting the `Object.prototype` which I dislike on principal, granted `should` probably doesn't give many problems....

jonasfj:
Okay, the resolvers/reapers... still have the potential timeout issues you mentioned. I think we should solve it at azure library level with timeouts... Note, https://github.com/gluwer/azure-table-node the library we use for table storage already defaults to 30s timeouts... IMO, we should still set this somewhere and make sure that `base.Entity` uses an HTTP agent... but that's all at `base.Entity` level patches suggestions are most welcome...

And implement producer consumer pattern for better more stable performance...

PR #33, opened 2015-02-19
Fix the pending tasks method to return 0 when there are no tasks

jhford:
instead of {}


jonasfj:
ahh... this probably a bug in some part of express or superagent that doesn't return the body because the body is "0" and thus it doesn't think there is one...

I'll add test case for this... It's fixed in the PR #34... Where I also changed the JSON structure for this API end-point... Just so we can extend it with additional properties in the future.
(It's always bad to return a list or number).

PR #34, opened 2015-02-20
Backport pollflow

jonasfj:
* Marked everything deprecated accordingly
  * Updated JSON schemas to match what we will have in the future

Shouldn't break compatibility, but allow migration to the new poll-flow, as documented:
http://docs.taskcluster.net/queue/worker-interaction/

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -14,7 +14,7 @@ mv /tmp/pg_hba.conf /etc/postgresql/9.1/main/pg_hba.conf;
 sudo service postgresql restart;
 
 # Dependencies that we are likely to change over time
-sudo apt-get update;
-sudo apt-get -y install screen vim;
+#sudo apt-get update;
+#sudo apt-get -y install screen vim;
 SCRIPT
 end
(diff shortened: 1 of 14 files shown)


jonasfj:
Note, don't worry about travis tests... they are horribly unstable..

jhford:
Looks good!

jonasfj:
@jhford,
So will you give me an `r+` on: https://bugzilla.mozilla.org/show_bug.cgi?id=1123994#c10

jonasfj:
@gregarndt, this is designed to backwards compatible with the current flow... And support the new flow, so we can transition slowly.
I'll deploy it when I have your `r+` on https://bugzilla.mozilla.org/show_bug.cgi?id=1123994#c10, then you can start testing out docker-worker AMIs against production queue, and switch workerTypes over the new AMI one by one. No need for hard switches.

@gregarndt, I'm not trying pressure you on docker-worker supporting new queue poll flow. But please let me know if anything blocks progress :)
As far as I can see we just need docker-worker to be ported + reviews and deployment (which will be "interesting").

PR #35, opened 2015-04-10
Fix node 0.12

jonasfj:
* Lots of packages upgraded
 * Works with node 0.12
 * Uses keepAlive for azure services (disables nagle too)
 * Load test utilities

Performance is ~100 req/s when running a 1X dyno... And ~120 req/s running a 2X dyno.
Before with node 0.10 not using keepAlive, we had 

**Before** node 0.10 no keepAlive:
 * `24.33 req / s` and 0 request timeouts (192 concurrent requests for 3 min)
 * `15.06 req / s` and 785 request timeouts (384 concurrent requests for 3 min)

**After** node 0.12 with keepAlive:
 * `104.09  req/s` and 0 request timeouts (192 concurrent requests for 3 min) 
 * `99.01  req/s` and 0 request timeouts (384 concurrent requests for 3 min)

We still slow down with too many requests... But it's less of a problem. I am tempted to use something like [toobusy](https://github.com/lloyd/node-toobusy) to measure event-loop lag and use this to reject incoming requests 503. But heroku router won't use this to try another dyno, granted it would make individual dynos less prone to melting. But ideally we should reject connections at TCP level, then heroku router will quarantine the node for 5s and try up to 10 other dynos. See heroku router docs...

Anyways, I suspect we should be able to over-provision so much now that we won't have problems. Our current peak is 5k req/s, otherwise we're mostly below 2-3k req/s, so we should be able to handle the current load with one or two dynos... But I say we splash a little and maybe use 4 :)


package.json
```diff
@@ -15,30 +15,32 @@
...
-    "azure-storage":        "0.4.2",
-    "debug":                "2.1.1",
-    "lodash":               "^3.0.0",
+    "azure-storage":        "0.4.3",
+    "debug":                "^2.1.3",
+    "lodash":               "^3.6.0",
     "promise":              "6.1.0",
     "slugid":               "1.0.3",
-    "taskcluster-base":     "0.7.15",
+    "taskcluster-base":     "taskcluster/taskcluster-base#c3b1647",
```
jonasfj: this is PR: https://github.com/taskcluster/taskcluster-base/pull/9

bin/load-test.js
```diff
@@ -0,0 +1,167 @@
...
+  summary();
+
+  console.log("Exiting");
+  exiting = true;
+};
+
+// If load-test.js is executed start the load-test
+if (!module.parent) {
+  // Find configuration profile
+  var profile = process.argv[2];
```
lightsofapollo-taskcluster: fwiw- there is no good reason to use var now that we can safely use let

queue/queueservice.js
```diff
@@ -88,20 +116,135 @@ class QueueService {
...
+              );
+            });
+          }
+        };
+      });
+    });
+  }
+
+  /////// START OF IMPLEMENTATION WITH AZURE STORAGE */
+/*
```
lightsofapollo-taskcluster: version control does a fine job of keeping history around if you really want to keep these commented in can you explain why (as a comment)

lightsofapollo-taskcluster:
@jonasfj lgtm... Though I am curious if you tested performance with and without babel compiled a head of time.

bin/load-test.js
jonasfj: yeah, lack of syntax highlighting is not the best reason to be using `var` over `let` :)

queue/queueservice.js
jonasfj: Good point... We need to use azure-storage for SAS... but otherwise we don't need it... Anyways, it's nice to keep the impl. that works with it around... now that we have it. So we can test with it over time.

jonasfj:
Update. for the record new heroku dynos gives ~93 req/s (ie. the new 1X)

package.json
```diff
@@ -3,11 +3,11 @@
...
-    "start":                "babel-node -r bin/server.js           $PROFILE",
-    "claim-reaper":         "babel-node -r bin/claim-reaper.js     $PROFILE",
-    "deadline-reaper":      "babel-node -r bin/deadline-reaper.js  $PROFILE",
-    "expire-tasks":         "babel-node -r bin/expire-tasks.js     $PROFILE",
-    "expire-artifacts":     "babel-node -r bin/expire-artifacts.js $PROFILE",
+    "start":                "babel-node bin/server.js           $PROFILE",
+    "claim-reaper":         "babel-node bin/claim-reaper.js     $PROFILE",
+    "deadline-reaper":      "babel-node bin/deadline-reaper.js  $PROFILE",
+    "expire-tasks":         "babel-node bin/expire-tasks.js     $PROFILE",
+    "expire-artifacts":     "babel-node bin/expire-artifacts.js $PROFILE",
```
jhford: Do we no longer need -r?

queue/queueservice.js
jhford: if it's living in a comment, it's not really around anymore.  without some sort of test or deployment, it's just going to end up confusing the person who finds it and wants to use it but doesn't realise that this code no longer is relevant.

test/api/artifact_test.js
```diff
@@ -16,6 +16,36 @@ suite('Post artifacts', function() {
...
+      res = err.response;
+    }
+    expect(res.statusCode).to.be(303);
+    console.log(res.headers.location);
+    return request.get(res.headers.location).end();
+  };
+
+  // Get something we expect to return 404, this is just easier than having
+  // try/catch blocks all over the code
+  var get404 = async (url) => {
```
jhford: good idea!

PR #36, opened 2015-05-01
Add support for proxing for EC2 regions, and prod config

jonasfj:
We probably need more testing...

Like testing the actual proxies... at the very least...

Probably we can do testing of EC2 region hacks... by using x-forwarded-for in unit tests.


jonasfj:
Tests pass locally, I think this is sufficient test to deploy it...

PR #38, opened 2015-05-17
Use  for exception run

jonasfj:
From [bug 1148965](https://bugzilla.mozilla.org/show_bug.cgi?id=1148965).

>   A) Set {reasonCreated: 'exception'}, if a run was created solely for one of the two cases:
     1) task canceled before first run was scheduled
     2) deadline expired before first run was schedule

### DIFF
--- a/queue/deadlineresolver.js
+++ b/queue/deadlineresolver.js
@@ -147,11 +147,14 @@ class DeadlineResolver {
 
     // Ensure that all runs are resolved
     await task.modify((task) => {
+      // If there is no run, we add a new one to signal that the task is
+      // resolved. As this run is purely to signal an exception, we set
+      // `reasonCreated: 'exception'`.
       if (task.runs.length === 0) {
         var now = new Date().toJSON();
         task.runs.push({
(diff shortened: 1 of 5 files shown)


petemoore:
Given that we already create a new run when a task is cancelled/expired before it is scheduled, I think this change to the reasonCreated totally makes sense. r+

On the separate topic of whether it makes sense to create a new run for a cancellation when a task has not yet been scheduled - I see pros and cons - the biggest con being that it seems a little unintuitive at first for someone fresh on the topic. After some reasoning, it makes more sense.

Partly this is probably wording ("run" is more like an event in the ledger of events that occurs but is maybe not ideally named). We discussed on IRC the possibility of recording a log of events, each being either a run or a cancel event, but overall I agree that the added complexity of introducing two types of events (a run or a task cancellation) is not outweighed by the advantages it brings for the few use cases it serves. So let's keep an eye on it, but leave it as it is. In this case probably the simplicity of having a set of runs, all in the same data structure, serves us well.

PR #39, opened 2015-05-26
Allow artifact creation past resolution as exception

jonasfj:
Allow artifact creation past task resolution if resolved as `exception`.

Generally, tasks resolved as exception aren't being resolved in a calm and orderly fashion, and any resulting artifact or log is a best-effort service.

### DIFF
--- a/routes/artifacts.js
+++ b/routes/artifacts.js
@@ -173,7 +173,12 @@ api.declare({
     }
     if (!allow) {
       return res.status(409).json({
-        message:    "The given run is not running",
+        message:    "Artifacts cannot be created for a task after it is " +
+                    "resolved, unless it is resolved 'exception', and even " +
+                    "in this case only up to 25 min past resolution." +
+                    "This to ensure that artifacts have been uploaded before " +
+                    "a task is 'completed' and output is consumed by a " +
+                    "dependent task",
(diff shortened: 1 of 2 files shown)


routes/artifacts.js
```diff
@@ -163,13 +163,23 @@ api.declare({
...
+    if (run.state === 'exception') {
+      // If task was resolved exception, we'll allow artifacts to be uploaded
+      // up to 25 min past resolution. This allows us to report exception as
+      // soon as we know and then upload logs if possible.
+      // Useful because exception usually implies something badly wrong.
+      allow = (new Date(run.resolved).getTime() > Date.now() - 25 * 60 * 1000);
+    }
+    if (!allow) {
+      return res.status(409).json({
+        message:    "The given run is not running",
```
gregarndt: Do you think this message should be more specific that it's not running and past the 25 minute window?

jonasfj: Well, it's a bit more complicated, because resolved as `completed` won't allow for upload 25min past resolution.

But yeah, making it a lot more explanatory can't hurt. Certainly not, I'll push a better messages. Thanks for pointing this out, we probably have a lot of error messages that could be more explanatory...

PR #40, opened 2015-05-28
Refactor test to use assume

jonasfj:
@jhford, hi is this entirely crazy.... (probably).

Review is not really needed here.. .Just take a peak and see if you agree that `assume` beats `expect.js` in terms of being more modern and `should.js` by not poisoning the prototype.

I plan to use `assume` more extensively, so this is just as much a sanity check to hear what you think of the module?

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
   "devDependencies": {
     "mocha":                "^2.2.1",
     "xmlbuilder":           "^2.6.2",
-    "expect.js":            "^0.3.1"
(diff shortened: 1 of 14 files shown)


jonasfj:
@jhford, I'm merging this not to block any further as I have other things I want to land.. .and I don't want this pending because it creates conflicts...

I'd love to still get some comments idea... feedback noise regarding the choice of `assume`. If I get none I'll start using it in other modules too...

PR #41, opened 2015-06-06
Fix queue collisions

jonasfj:
This address two bugs:
https://bugzilla.mozilla.org/show_bug.cgi?id=1170821
https://bugzilla.mozilla.org/show_bug.cgi?id=1162943

Once this is rolled out, we give it a week or a few days, and then remove all the legacy queues. So that pull URLs again returns an array of one queue.

For now we abuse the list returned for priority, to migrate underlying queue names to a new name without dropping tasks.

### DIFF
--- a/queue/queueservice.js
+++ b/queue/queueservice.js
@@ -392,10 +392,14 @@ class QueueService {
     }
   }
 
-  /** Remove all worker queues not used since `now - 10 days` */
+  /**
+   * Remove all worker queues not used since `now - 10 days`.
+   * Returns number of queues deleted.
+   */
   async deleteUnusedWorkerQueues(now = new Date()) {
     assert(now instanceof Date, "Expected now as Date object");
     let deleteIfNotUsedSince = now.getTime() - 10 * 24 * 60 * 60 * 1000;
+    let deleted = 0; // Number of queues deleted
 
     // Iterate through all pages
     let marker = undefined;
@@ -431,11 +435,17 @@ class QueueService {
         }
 
         debug("Deleting queue %s with metadata: %j", name, metadata);
-        return this.client.deleteQueue(name);
+        await this.client.deleteQueue(name);
+
+        // Count queues deleted (for test ability)
+        deleted += 1;
       }));
 
       // Keep going until we get an `undefined` marker
     } while (marker);
+
+    // Return number of queues deleted
+    return deleted;
   }
 
 
--- a/queue/queueservice.js
+++ b/queue/queueservice.js
@@ -316,8 +316,8 @@ class QueueService {
 
     // Return and cache promise that we created this queue
     return this.queues[id] = Promise.all([
-      this._ensureQueueAndMetadata(name),
-      this._ensureQueueAndMetadata(legacyName)
+      this._ensureQueueAndMetadata(name, provisionerId, workerType),
+      this._ensureQueueAndMetadata(legacyName, provisionerId, workerType)
     ]).catch(err => {
       debug("[alert-operator] Failed to ensure azure queue: %s, as JSON: %j",
             err, err, err.stack);
@@ -344,9 +344,29 @@ class QueueService {
     // Fetch meta-data from queue, checking if it exists
     try {
       let {metadata} = await this.client.getMetadata(queue);
+
+      // Check if meta-data is up-to-date
+      var lastUsed = new Date(metadata.last_used);
+      if (metadata.provisioner_id === provisionerId
+          && metadata.worker_type === workerType
+          && isFinite(lastUsed)
+          && lastUsed.getTime() > Date.now() - 23 * 60 * 60 * 1000
+      ) {
+        return; // We're done as meta-data is present
+      }
+
+      // Update meta-data
+      return this.client.setMetadata(queue, {
+        provisioner_id: provisionerId,
+        worker_type:    workerType,
+        last_used:      taskcluster.fromNowJSON()
+      });
     } catch (err) {
-      // We handle queue not found exceptions
-      if (err.code !== 'QueueNotFound') {
+      // We handle queue not found exceptions, because getMetadata is a HEA
+      // request we don't get any error message payload, so we also accept 404
+      // as implying the same.
+      if (err.code !== 'QueueNotFound' &&
+          err.statusCode !== 404) {
         throw err;
       }
 
@@ -369,27 +389,7 @@ class QueueService {
           throw err;
         }
       }
-
-      // Queue created (with meta-data) and we're done
-      return;
     }
-
-    // Check if meta-data is up-to-date
-    var lastUsed = new Date(metadata.last_used);
-    if (metadata.provisioner_id === provisionerId
-        && metadata.worker_type === workerType
-        && isFinite(lastUsed)
-        && lastUsed.getTime() > Date.now() - 23 * 60 * 60 * 1000
-    ) {
-      return; // We're done as meta-data is present
-    }
-
-    // Update meta-data
-    await this.client.setMetadata(queue, {
-      provisioner_id: provisionerId,
-      worker_type:    workerType,
-      last_used:      taskcluster.fromNowJSON()
-    });
   }
 
   /** Remove all worker queues not used since `now - 10 days` */
@@ -516,8 +516,8 @@ class QueueService {
 
     // For each queue name, return signed URLs for the queue
     var queues = [
-      queueName,
-      legacyName
+      legacyName, // Return legacy queue first, so we it emptied
+      queueName
     ].map(queueName => {
       // Create shared access signature
       var sas = this.client.sas(queueName, {
@@ -566,9 +566,6 @@ class QueueService {
         .replace(/=*$/, ''),
       '1'             // priority, currently just hardcoded to 1
     ].join('-');
-
-    // Get queue meta-data
-    return this._countMessages(queueName);
 
     // Find messages count from primary and legacy queue
     var [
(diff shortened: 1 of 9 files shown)


queue/queueservice.js
```diff
@@ -288,8 +284,24 @@ class QueueService {
...
+      var h = crypto.createHash('sha256').update(id).digest();
+      return base32.encode(h.slice(0, 15)).toString('utf-8').toLowerCase();
+    };
+
     // Construct queue name
     var name = [
+      this.prefix,            // prefix all queues
+      hashId(provisionerId),  // hash of provisionerId
+      hashId(workerType),     // hash of workerType
+      '1'                     // priority, currently just hardcoded to 1
```
gregarndt: Not sure what priority is used for in the queue name, but I see that both the new queue and legacy queue have the same priority....

queue/queueservice.js
```diff
@@ -303,15 +315,140 @@ class QueueService {
...
+      debug("[alert-operator] Failed to ensure azure queue: %s, as JSON: %j",
+            err, err, err.stack);
+
       // Don't cache negative results
       this.queues[id] = undefined;
       throw err;
-    }).then(() => {
-      return name;
-    });
+    }).then(() => name);
```
gregarndt: Do we only care about returning the name and not both name and legacyname?

queue/queueservice.js
```diff
@@ -303,15 +315,140 @@ class QueueService {
...
+          worker_type:    workerType,
+          last_used:      taskcluster.fromNowJSON()
+        });
+      } catch (err) {
+        // If queue already exists, we must have been racing we assume meta-data
+        // is up to date...
+        if (err.code !== 'QueueAlreadyExists') {
+          // We probably don't have to alert on all these. But we should
+          // definitely alert if we see a QueueBeingDeleted error
+          debug("[alert-operator] Failed to create queue: %s for %s/%s " +
```
gregarndt: Something minor but template strings makes this a little nicer to read :)

queue/queueservice.js
```diff
@@ -303,15 +315,140 @@ class QueueService {
...
+      let {metadata} = await this.client.getMetadata(queue);
+
+      // Check if meta-data is up-to-date
+      var lastUsed = new Date(metadata.last_used);
+      if (metadata.provisioner_id === provisionerId
+          && metadata.worker_type === workerType
+          && isFinite(lastUsed)
+          && lastUsed.getTime() > Date.now() - 23 * 60 * 60 * 1000
+      ) {
+        return; // We're done as meta-data is present
```
gregarndt: Is there something elsewhere that sets last used?  If we are ensuring the queue exists, does that mean we're intending to use it so last used is now?

gregarndt: Nevermind, I see the comment that we tolerate a 23 hour difference.

queue/queueservice.js
```diff
@@ -303,15 +315,140 @@ class QueueService {
...
+          || lastUsed.getTime() < deleteIfNotUsedSince
+        );
+      });
+
+      // Delete queues, if they are empty
+      await Promise.all(queues.map(async ({name, metadata}) => {
+        // Fetch message count (approximate)
+        let {messageCount} = await this.client.getMetadata(name);
+        if (messageCount > 0) {
+          return; // Abort if there is messages
```
gregarndt: nit: "if there are messages" :)

queue/queueservice.js
```diff
@@ -357,7 +494,7 @@ class QueueService {
   }
 
   /** Get signed URLs for polling and deleting from the azure queue */
-  async signedPendingPollUrl(provisionerId, workerType) {
+  async signedPendingPollUrls(provisionerId, workerType) {
     // Find name of azure queue
     var queueName = await this.ensurePendingQueue(provisionerId, workerType);
```
gregarndt: Ok, I see earlier why returning just the queue name is easier here, but won't there be more than 1 pending queue?

queue/queueservice.js
```diff
@@ -369,30 +506,55 @@ class QueueService {
...
+      base32.encode(decodeUrlSafeBase64(workerType))
+        .toString('utf8')
+        .toLowerCase()
+        .replace(/=*$/, ''),
+      '1'             // priority, currently just hardcoded to 1
+    ].join('-');
+
+    // For each queue name, return signed URLs for the queue
+    var queues = [
+      legacyName, // Return legacy queue first, so we it emptied
```
gregarndt: comment seems weird  probably should be "so we empty it"

queue/queueservice.js
```diff
@@ -400,8 +562,31 @@ class QueueService {
...
+      this.prefix,    // prefix all queues
+      base32.encode(decodeUrlSafeBase64(provisionerId))
+        .toString('utf8')
+        .toLowerCase()
+        .replace(/=*$/, ''),
+      base32.encode(decodeUrlSafeBase64(workerType))
+        .toString('utf8')
+        .toLowerCase()
+        .replace(/=*$/, ''),
+      '1'             // priority, currently just hardcoded to 1
```
gregarndt: nit: looks like indent got messed up here.  I know how you like your indent.

queue/queueservice.js
```diff
@@ -400,8 +562,31 @@ class QueueService {
...
+    ].join('-');
+
+    // Find messages count from primary and legacy queue
+    var [
+      count,
+      legacyCount
+    ] = (await Promise.all([
```
gregarndt: So I know for now we're operating only on one main queue (2 while transitioning between legacy to the new naming) , but at some point I know we spoke of supporting queue priority.  Hardcoding the queues here for counts seems to make it rigid with the queues that are supported.  Shouldn't this iterate over something returned from listQueues or similar?

test/queue/queueservice_test.js
```diff
@@ -164,4 +171,70 @@ suite('queue/QueueService', function() {
...
+    let workerType = slugid.v4();
+    let queueName = await queueService.ensurePendingQueue(
+      provisionerId, workerType
+    );
+
+    // Let's delete from now and check that we didn't delete queue just created
+    await queueService.deleteUnusedWorkerQueues();
+
+    // Get meta-data, this will fail if the queue was deleted
+    await queueService.client.getMetadata(queueName);
```
gregarndt: I haven't run the tests, but will this just show that there was an exception thrown?  Would it be better to catch the exception and give a reason why this failed as an assertion?

test/queue/queueservice_test.js
```diff
@@ -164,4 +171,70 @@ suite('queue/QueueService', function() {
...
     assert(typeof(count) === 'number', "Expected count as number!");
   });
+
+  test("deleteUnusedWorkerQueues (can delete queues)", async () => {
+    // 11 days into the future, so we'll delete all queues (yay)
+    let now = new Date(Date.now() + 11 * 24 * 60 * 60 * 1000);
+    let deleted = await queueService.deleteUnusedWorkerQueues(now);
+    // There should always be at least one queue, because test cases above
+    // will have created one we can delete...
+    assume(deleted).is.atleast(1);
```
gregarndt: so this is checking that the value returned is at least 1, but does not check that the queue indeed was deleted.  Could we then try to get the meta data on the queue and ensure that throws an exception?

test/queue/queueservice_test.js
```diff
@@ -164,4 +171,70 @@ suite('queue/QueueService', function() {
     debug("pending message count: %j", count);
     assert(typeof(count) === 'number', "Expected count as number!");
   });
+
+  test("deleteUnusedWorkerQueues (can delete queues)", async () => {
+    // 11 days into the future, so we'll delete all queues (yay)
+    let now = new Date(Date.now() + 11 * 24 * 60 * 60 * 1000);
+    let deleted = await queueService.deleteUnusedWorkerQueues(now);
+    // There should always be at least one queue, because test cases above
```
gregarndt: So.. this is a nit at least for me, but relying on data to be setup by previous tests makes it difficult to run this specific test in isolation and depends on the sequence of other tests.

queue/queueservice.js
jonasfj: The `priority` part of the name is just so we create new queues in the future...

Ie. when we decide to add priority support we add a queue where the priority number is `2` and `3`, and then we have 3 queue names... Hence, we can implement priority...

queue/queueservice.js
jonasfj: Oh, so instead of having the `+` ?

Doing `debug('[alert-operator] Failed to create queue: ${queue} '...)` is a lot less efficient than `debug('[alert-operator] Failed to create queue: %s '..., queue)`, because `debug` doesn't do the string formatting if `DEBUG` isn't set... Then the `debug` function is just an empty function if the particular file has been disabled, hence, no overhead.

queue/queueservice.js
jonasfj: Sure, let's care about priority when someone needs it...

queue/queueservice.js
jonasfj: Perhaps we should change `ensurePendingQueue` to return a promise for an array of queue names ordered by priority... but for now, this is a step in the right direction. If you define the concept priority I'll happily implement it :)

jonasfj: it should be a lot simpler since the `signedPendingPollUrls` now returns a list :)

test/queue/queueservice_test.js
jonasfj: I generally agree... but we have dirty state the cloud here.. so nothing we do is nice :)

test/queue/queueservice_test.js
jonasfj: Sure, if I'm willing to pay the cost when/if this test was to fail... I suspect it won't fail anytime soon.

test/queue/queueservice_test.js
jonasfj: good point...

PR #42, opened 2015-07-20
Removed legacy queue creation

jonasfj:
Remove legacy queue names from bug 1170821...


gregarndt:
If we don't need these legacy queues any longer, seems awesome that we're getting rid of them.

PR #43, opened 2015-07-21
Task priority

jonasfj:
https://bugzilla.mozilla.org/show_bug.cgi?id=1178877

### DIFF
--- a/queue/queueservice.js
+++ b/queue/queueservice.js
@@ -42,10 +42,17 @@ var validateTask = function(task) {
 
 /** Priority to constant for use in queue name (should be a string) */
 const PRIORITY_TO_CONSTANT = {
-  normal: '1',
-  high:   '5'
+  high:   '5',
+  normal: '1'
 };
 _.forIn(PRIORITY_TO_CONSTANT, v => assert(typeof v === 'string'));
+
+/** Priority in order of priority from high to low */
+const PRIORITIES = [
+  'high',
+  'normal'
+];
+assert(_.xor(PRIORITIES, _.keys(PRIORITY_TO_CONSTANT)).length === 0);
 
 /**
  * Wrapper for azure queue storage, to ease our use cases.
@@ -505,7 +512,10 @@ class QueueService {
     var pendingPollTimeout = Math.floor(this.pendingPollTimeout / 1000);
 
     // For each queue name, return signed URLs for the queue
-    let queues = _.map(queueNames, queueName => {
+
+    let queues = PRIORITIES.map(priority => {
+      let queueName = queueNames[priority];
+
       // Create shared access signature
       var sas = this.client.sas(queueName, {
         start, expiry,
--- a/queue/queueservice.js
+++ b/queue/queueservice.js
@@ -40,6 +40,13 @@ var validateTask = function(task) {
   assert(task.deadline instanceof Date, "Expected task.deadline");
 };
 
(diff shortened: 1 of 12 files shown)


queue/queueservice.js
```diff
@@ -40,6 +40,20 @@ var validateTask = function(task) {
...
 
+/** Priority to constant for use in queue name (should be a string) */
+const PRIORITY_TO_CONSTANT = {
+  high:   '5',
+  normal: '1'
+};
+_.forIn(PRIORITY_TO_CONSTANT, v => assert(typeof v === 'string'));
+
+/** Priority in order of priority from high to low */
+const PRIORITIES = [
```
gregarndt: Why not just use this both for the list of priorities and the constant they map to?

queue/queueservice.js
```diff
@@ -291,25 +306,28 @@ class QueueService {
...
 
-    // Construct queue name
-    var name = [
+    // Construct queue name prefix (add priority later)
+    let namePrefix = [
       this.prefix,            // prefix all queues
       hashId(provisionerId),  // hash of provisionerId
       hashId(workerType),     // hash of workerType
-      '1'                     // priority, currently just hardcoded to 1
+      ''                      // priority, add 1 = normal, 5 = high
```
gregarndt: Is this just so there is documentation as to what the last value of the string will be?

queue/queueservice.js
```diff
@@ -468,7 +486,7 @@ class QueueService {
     }
 
     // Put message queue
-    return this._putMessage(queueName, {
+    return this._putMessage(queueNames[task.priority], {
```
gregarndt: isn't queueNames:

```js
{
  <provisionerId>/<workerType>: [<normal priority>, <high priority>]
}
```
If so, the keys in queueNames wouldn't be task.priority... unless I'm misunderstanding something

schemas/create-task-request.yml
```diff
@@ -63,6 +63,16 @@ properties:
       minLength:    1
     maxItems:       10
     uniqueItems:    true
+  priority:
+    title:          "Task Priority"
+    description: |
+      Priority of task, this defaults to `normal` and the scope
+      `queue:task-priority:high` is required to define a task with priority
+      to `high`. Additional priority may be added later.
```
gregarndt: nit: of `high`

queue/queueservice.js
jonasfj: Well... Then the constant becomes implicit, ie. it becomes the index of the element in the array...

I think it would work, even look nice, the downside is that if we change things here add/remove a priority level we really have to keep using the old constants :)
Or write some code to keep legacy constants around...

If accidentally the constant that is used for a level changes, then all pending tasks might be lost... I would argue that sucks :)
So keeping them explicit makes it harder to make mistakes.

queue/queueservice.js
jonasfj: Yeah, and to make sure that the string ends with a `-`.

schemas/create-task-request.yml
jonasfj: Or `set to high`, but yeah... Will fix that, thanks :)

gregarndt:
Looks good to me except for the test failing on Travis.  I assume the other tests just work without setting priority because it's defaulted to normal.

jonasfj:
Yeah, travis tests are just stupid... it all works locally. Need to port to circle ci..

PR #44, opened 2015-07-24
Bug 1184272: specify all API scopes as [[str]]

djmitche:


jonasfj:
Looks right to me... merged...

PR #45, opened 2015-07-28
Fixed scopes

jonasfj:
This should fix some issues with previous patch that I should have caught in review... or automated testing, unfortunately I'm human and travis is too buggy to trust...


djmitche:
Ugh, yes, sorry.  Better testing++

PR #48, opened 2015-07-30
Error response should include artifact details

gregarndt:
When a request is made to update an artifact, details about the original
as well as the new artifact should be included in the error response
when the artifact can not be updated.

### DIFF
--- a/test/api/artifact_test.js
+++ b/test/api/artifact_test.js
@@ -408,7 +408,7 @@ suite('Post artifacts', function() {
       workerId:       'my-worker'
     });
 
-    debug("### Report exception");
+    debug("### Report completed");
     await helper.queue.reportCompleted(taskId, 0);
 
     debug("### Send post artifact request");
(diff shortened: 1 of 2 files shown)


test/api/artifact_test.js
```diff
@@ -445,4 +445,88 @@ suite('Post artifacts', function() {
...
+    debug("### Send second post artifact request to update expiration");
+    await helper.queue.createArtifact(taskId, 0, 'public/s3.json', {
+      storageType:  's3',
+      expires:      expirationIn2Days,
+      contentType:  'application/json'
+    }).catch(async (err) => {
+      await helper.queue.reportCompleted(taskId, 0);
+      debug("Got error: %s, as JSON %j", err, err);
+      assume().fail("Expected artifact to be updated");
+    });
```
jonasfj: No need to catch the error just to throw it again...

test/api/artifact_test.js
```diff
@@ -445,4 +445,88 @@ suite('Post artifacts', function() {
...
+
+    var artifacts = await helper.queue.listArtifacts(taskId, 0);
+
+    await helper.queue.reportCompleted(taskId, 0);
```
jonasfj: No need to complete the task... just leave it as it is... this is tests...

test/api/artifact_test.js
```diff
@@ -445,4 +445,88 @@ suite('Post artifacts', function() {
...
+    }).then(async () => {
+      await helper.queue.reportCompleted(taskId, 0);
+      assume().fail("Expected request to be unsuccessful");
+    }, async (err) => {
+      debug("Got error: %s, as JSON %j", err, err);
+      await helper.queue.reportCompleted(taskId, 0);
+      assume(err.message).includes('Artifact already exists');
+      assume(err.body.error).contains('originalArtifact', 'newArtifact');
+    });
+
```
jonasfj: Can't we have reportCompelted here...

gregarndt: So my main reasoning for reporting completed is to just clean up after ourselves in a test.  I did not know how strict you are with this, but I try to do my best at it if I remember to.  It's not completely necessary, but I'm not sure it's a problem doing so.  Here I do it before I make any assertions because won't it stop when the first assertion/assume fails and not execute anything after that?

jonasfj:
Minor things... just suggestions to keep tests simple... but generally I'm okay if you just want to land this.

test/api/artifact_test.js
jonasfj: true  it will... and we don't care about clean up after tests... not for things in azure table storage... we just leave them there forever :)

test/api/artifact_test.js
gregarndt: addressed in e7a6246

jonasfj:
@gregarndt, nvm, these nits... I'll take a look at it as I merge it in...

PR #50, opened 2015-08-01
S3 artifact headers

jonasfj:
https://bugzilla.mozilla.org/show_bug.cgi?id=1190016

### DIFF
--- a/queue/bucket.js
+++ b/queue/bucket.js
@@ -52,7 +52,9 @@ Bucket.prototype.createPutUrl = function(prefix, options) {
     that.s3.getSignedUrl('putObject', {
       Key:          prefix,
       ContentType:  options.contentType,
-      Expires:      options.expires
+      Expires:      options.expires,
+      ContentMD5:   options.contentMD5,
+      Metadata:     options.metadata
     }, function(err, url) {
       if (err) {
         return reject(err);
--- a/routes/artifacts.js
+++ b/routes/artifacts.js
@@ -190,6 +190,18 @@ api.declare({
         details.bucket  = this.privateBucket.bucket;
       }
       details.prefix    = [taskId, runId, name].join('/');
+      if (input.contentMD5 || input.contentSHA256) {
+        // Remove this check, when contentMD5 and contentSHA256 are required
+        if (!input.contentMD5 || !input.contentSHA256) {
+          return res.status(400).json({
+            message:  "Both contentMD5 and contentSHA256 must be given " +
+                      "if either is given"
+          });
+        }
+        // Add contentMD5 and contentSHA256 to details
+        details.contentMD5    = input.contentMD5;
+        details.contentSHA256 = input.contentSHA256;
+      }
       break;
 
     case 'azure':
@@ -296,17 +308,30 @@ api.declare({
         bucket = this.privateBucket;
       }
       // Create put URL
+      var metadata = {};
+      if (artifact.details.contentSHA256) {
+        metadata['content-sha256'] = artifact.details.contentSHA256;
+      }
       var putUrl = await bucket.createPutUrl(
         artifact.details.prefix, {
         contentType:      artifact.contentType,
-        expires:          30 * 60 + 10 // Add 10 sec for clock drift
+        expires:          30 * 60 + 10, // Add 10 sec for clock drift
+        contentMD5:       artifact.details.contentMD5 || undefined,
+        metadata:         metadata
       });
-      return res.reply({
+      // Construct reply
+      var reply = {
         storageType:  's3',
         contentType:  artifact.contentType,
         expires:      expiry.toJSON(),
-        putUrl:       putUrl
-      });
+        putUrl:       putUrl,
+        cacheControl: (isPublic ? 'public' : 'private') + ', no-transform'
+      };
+      // Provide contentMD5 if present
+      if (artifact.details.contentMD5) {
+        reply.contentMD5 = artifact.details.contentMD5;
+      }
+      return res.reply(reply);
 
     case 'azure':
       // Reply with SAS for azure
--- a/schemas/constants.js
+++ b/schemas/constants.js
@@ -84,4 +84,8 @@ module.exports = {
 
   // Slugid pattern, for when-ever that is useful
   "slugid-pattern":  "^[a-zA-Z0-9-_]{22}$",
+
+  // Pattern for base64 encoded strings
+  "base64-pattern":  "^(?:[A-Za-z0-9+/]{4})*" +
+                     "(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)$"
 };
--- a/schemas/post-artifact-request.yml
+++ b/schemas/post-artifact-request.yml
@@ -31,11 +31,35 @@ oneOf:
            always using `application/json` for JSON artifacts.
         type:       string
         maxLength:  255
+      contentMD5:
+        description: |
+          Base64 encoded MD5 sum of the artifact. If specified, it **must**
+          also be given as value to the `Content-MD5` header when uploading
+          to S3.
+          **Warning**, this will likely be required in the future, and **must**
+          be specified if `contentSHA256` is specified.
+        type:       string
+        pattern:    {$const: base64-pattern}
+        minLength:  24
+        maxLength:  24
+      contentSHA256:
+        description: |
+          Base64 encoded SHA-256 sum of the artifact. If specified, it will
+          be available as `x-amz-meta-content-sha256` header when people
+          download the artifact. This is useful for validating integrity.
+          **Warning**, this will likely be required in the future, and **must**
+          be specified if `contentMD5` is specified.
+        type:       string
+        pattern:    {$const: base64-pattern}
+        minLength:  44
+        maxLength:  44
     additionalProperties: false
     required:
       - storageType
       - expires
       - contentType
+      # contentMD5 and contentSHA256 aren't required YET
+      # We most port all workers to provide these before we make them required!
   - title:          "Azure Artifact Request"
     description: |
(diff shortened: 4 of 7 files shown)


queue/bucket.js
```diff
@@ -52,7 +52,9 @@ Bucket.prototype.createPutUrl = function(prefix, options) {
     that.s3.getSignedUrl('putObject', {
       Key:          prefix,
       ContentType:  options.contentType,
-      Expires:      options.expires
+      Expires:      options.expires,
+      ContentMD5:   options.contentMD5,
+      Metadata:     options.metadata
     }, function(err, url) {
```
petemoore: `options.contentMD5` and `options.metadata` missing from doc comments (around line 43)

routes/artifacts.js
```diff
@@ -190,6 +190,18 @@ api.declare({
         details.bucket  = this.privateBucket.bucket;
       }
       details.prefix    = [taskId, runId, name].join('/');
+      if (input.contentMD5 || input.contentSHA256) {
+        // Remove this check, when contentMD5 and contentSHA256 are required
+        if (!input.contentMD5 || !input.contentSHA256) {
+          return res.status(400).json({
+            message:  "Both contentMD5 and contentSHA256 must be given " +
+                      "if either is given"
+          });
```
petemoore: It would be useful to include in the error message which hash was given (md5 or sha256) and which was not, and also other useful instance data for pinning down problem when troubleshooting (taskId, runId, name).

For example, the error message could be `You tried to upload artifact <name> from taskId <taskId> in run <runId>. You supplied a <hashSupplied> hash, but not a <hashNotSupplied> hash. You need to supply either both or neither.`

routes/artifacts.js
```diff
@@ -296,17 +308,30 @@ api.declare({
...
         expires:      expiry.toJSON(),
-        putUrl:       putUrl
-      });
+        putUrl:       putUrl,
+        cacheControl: (isPublic ? 'public' : 'private') + ', no-transform'
+      };
+      // Provide contentMD5 if present
+      if (artifact.details.contentMD5) {
+        reply.contentMD5 = artifact.details.contentMD5;
+      }
```
petemoore: should we also provide sha256?

schemas/constants.js
```diff
@@ -84,4 +84,8 @@ module.exports = {
 
   // Slugid pattern, for when-ever that is useful
   "slugid-pattern":  "^[a-zA-Z0-9-_]{22}$",
+
+  // Pattern for base64 encoded strings
```
petemoore: Interesting, I didn't realise base64 strings were always a multiple of 4 characters. Maybe worth adding a comment to explain that they are a multiple of 4 characters, with either one or two `=` characters as RHS padding.

schemas/constants.js
```diff
@@ -84,4 +84,8 @@ module.exports = {
...
+  "base64-pattern":  "^(?:[A-Za-z0-9+/]{4})*" +
+                     "(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)$"
 };
```
petemoore: Interesting, I didn't realise base64 strings were always a multiple of 4 characters. Maybe worth adding a comment to explain that they are a multiple of 4 characters, with either one or two = characters as RHS padding.

schemas/post-artifact-request.yml
```diff
@@ -31,11 +31,35 @@ oneOf:
...
         type:       string
         maxLength:  255
+      contentMD5:
+        description: |
+          Base64 encoded MD5 sum of the artifact. If specified, it **must**
+          also be given as value to the `Content-MD5` header when uploading
+          to S3.
+          **Warning**, this will likely be required in the future, and **must**
+          be specified if `contentSHA256` is specified.
+        type:       string
```
petemoore: This applies to the entire PR, but adding it here. Is there a reason we prefer to base64 decode the MD5/SHA256 rather than leave them in their _already ascii_ form?

test/api/artifact_test.js
```diff
@@ -168,6 +169,114 @@ suite('Post artifacts', function() {
...
+    } catch (err) {
+      assume(err.response.statusCode).is.eql(400);
+    }
+
+    debug("### Expire artifacts");
+    // config/test.js hardcoded to expire artifact 4 days in the future
+    await helper.expireArtifacts();
+  });
+
   test("Post S3 artifact (with bad scopes)", async () => {
```
petemoore: tests++

routes/artifacts.js
```diff
@@ -190,6 +190,18 @@ api.declare({
         details.bucket  = this.privateBucket.bucket;
       }
       details.prefix    = [taskId, runId, name].join('/');
+      if (input.contentMD5 || input.contentSHA256) {
```
gregarndt: perhaps both of these should be defined as optional headers in the description above and then we can change it to "*must*" when it's required.

routes/artifacts.js
```diff
@@ -296,17 +308,30 @@ api.declare({
...
+      var metadata = {};
+      if (artifact.details.contentSHA256) {
+        metadata['content-sha256'] = artifact.details.contentSHA256;
+      }
       var putUrl = await bucket.createPutUrl(
         artifact.details.prefix, {
         contentType:      artifact.contentType,
-        expires:          30 * 60 + 10 // Add 10 sec for clock drift
+        expires:          30 * 60 + 10, // Add 10 sec for clock drift
+        contentMD5:       artifact.details.contentMD5 || undefined,
```
gregarndt: Wouldn't this either be defined or undefined?  Or are you doing this in case contentMD5 is an empty string?

schemas/post-artifact-request.yml
```diff
@@ -31,11 +31,35 @@ oneOf:
...
+        pattern:    {$const: base64-pattern}
+        minLength:  44
+        maxLength:  44
     additionalProperties: false
     required:
       - storageType
       - expires
       - contentType
+      # contentMD5 and contentSHA256 aren't required YET
+      # We most port all workers to provide these before we make them required!
```
gregarndt: nit: 'must'

schemas/post-artifact-response.yml
```diff
@@ -21,6 +21,10 @@ oneOf:
           URL to which a `PUT` request can be made to upload the artifact
           requested. Note, the `Content-Length` must be specified correctly,
           and the `ContentType` header must be set the value specified below.
+          Addtionally, you may specify `Content-Encoding: gzip` if the
```
gregarndt: nit: Additionally

routes/artifacts.js
```diff
@@ -190,6 +190,18 @@ api.declare({
         details.bucket  = this.privateBucket.bucket;
       }
       details.prefix    = [taskId, runId, name].join('/');
+      if (input.contentMD5 || input.contentSHA256) {
+        // Remove this check, when contentMD5 and contentSHA256 are required
+        if (!input.contentMD5 || !input.contentSHA256) {
```
gregarndt: Might be a stupid simple test, but any value in having a test to make sure both are present?

jonasfj: Basically, I see providing them as using the new API.
So either you use the new API correctly, or you tolerate your legacy request.

routes/artifacts.js
jonasfj: `contentMD5` schema ensures it can't be an empty string... So this is just for clarity... Yeah, maybe I should just delete it...

schemas/post-artifact-request.yml
jonasfj: What is their "ascii form" are you talking `hex`?

This an encoding of the binary format in base64. For `contentMD5` s3 expects `content-md5` header to be base64 encoded md5 sum. I figured doing the same with sha256 made sense...
It doesn't really matter, base64 is a bit shorter that's all... But we could make the sha256 sum as `hex`.

petemoore: No worries, this was my misunderstanding, I thought the hex string was being base64 encoded, I didn't realise the binary hash value was being base64 encoded. Since shas and md5s are traditionally represented in hex, if it wasn't for s3 expecting it in base64, i would have found it more conventional to represent in hex. That said, could we convert between base64 and hex, so that client-interfacing values are in hex for both md5 and sha256, even if we pass through to s3 in base64? This just seems more canonical. Unless clients handle the signed urls and need to set s3 headers themselves, no point forcing them to use both.

Probably if one is base64 encoded, we should no doubt base64 encode both for consistency.

petemoore: If the queue returns the headers as per your proposal in https://bugzilla.mozilla.org/show_bug.cgi?id=1190016#c5 section C) then I think it would be nice for the `createArtifact` (or `defineArtifact` if we decide to rename it) to receive both the hashes in hex for the sake of conventionality (e.g. command line tools typically report in hex, so it is the one people are most familiar with when troubleshooting/double checking the hash, etc).

petemoore:
@jonasfj I still think this is a useful feature to have, but after almost a year I'm hoping it hasn't bit rotted too much. Would you be able to look at the outstanding review comments from @gregarndt and me?

jonasfj:
Basically, this is on ice for now... There are other things we want to add like.. Ensuring we have file size, and that artifacts are created...

Hmm... Maybe we do need to get back to this though as we can probably do this as an first step...

petemoore:
Yeah, I still think it has value - especially hashsum validation...

jonasfj:
Agreed, I'm just saying this probably has to be done as part of a slightly bigger re-think of artifacts...

gregarndt:
@jhford @jonasfj It looks like this is a very old PR that might be replaced by some of the validation work that John is proposing for the new artifact api.  Can we confirm that this no longer needs to be open and close if necessary?

PR #51, opened 2015-08-24
Fix extra exception messages after deadline

jonasfj:
Avoid duplicate messages for exceptions after deadline... We should only report the messages if `deadline-exceeded` is the reason.

If it was cancelled, or something else... we should publish another message about that.

----
Note, this is not a violation of our promises... messages are at-least once, so delivering them an extra time a few hours later is technically okay (but not elegant). And in practice we consider it a bug.


bhearsum:
Seems reasonable? But I can't do more than rubberstamp+ it...

gregarndt:
LGTM

PR #52, opened 2015-08-28
Bug 1199634 - delete Vagrantfile

petemoore:
apt config needed fixing


jonasfj:
Instead let's just delete the vagrant env... it's not needed anymore... sorry.

PR #54, opened 2015-09-04
Remote signature validation

jonasfj:
Similar to the one for index...

Note: I'll upgrade tc-base to next version when https://github.com/taskcluster/taskcluster-base/pull/30 lands... It's a pure upgrade no code changes... so I won't file a PR for it...


jonasfj:
Note, don't expect travis to be succesful...

PR #57, opened 2015-10-14
Support for public bucket cdn

jonasfj:


queue/bucket.js
```diff
@@ -13,13 +13,21 @@ var Promise     = require('promise');
...
  */
 var Bucket = function(options) {
   assert(options,             "options must be given");
   assert(options.bucket,      "bucket must be specified");
   assert(options.credentials, "credentials must be specified");
+  assert(!options.bucketCDN || typeof(options.bucketCDN) === 'string',
+         "Expected bucketCDN to be a hostname or empty string for none");
+  if (options.bucketCDN) {
+    assert(/^https?:\/\//.test(options.bucketCDN), "bucketCDN must be http(s)");
+    assert(/[^\/]$/.test(options.bucketCDN),
```
gregarndt: you could always just remove the trailing slash and not even care after that.

jonasfj: testing for valid urls would be more sane... Remove slash is bad, because it could be that someone really wanted "http://myhost.com/folder//<prefix>", so I have code saying that would be bad :)

This is really just being defensive...

jonasfj: "https://hostname.com" or "hostname.com" is easy to get wrong... as is whether or not to have trailing slash.

PR #58, opened 2015-10-19
Bug 1193607: update schema for limite scope charset

djmitche:
I can make similar changes for all components -- but would it be better to define a scope (and scopelist, scopeset, etc.) in taskcluster-base, and refer to that schema from elsewhere?


jonasfj:
I like this... I don't like referencing external schemas when it's not strictly necessary. For big things, I like it... For small constants not so much... It does sometimes have some quirks and this case could even have security implications...

I think we should add the pattern to `schema/constants.js` and then maybe have a file with constants in tc-base that is shared between all components... That would be very nice at we could use the same pattern for taskId and other things all over...

djmitche:
Sounds good.  How's this?

jonasfj:
the first one was good enough... this is just even better...

So just to be clear, `r+`.

PR #59, opened 2015-10-26
add trailing `$` to scope pattern

djmitche:


jonasfj:
I think we finally got this right... :)

PR #60, opened 2015-10-27
Bug 1218913 - added missing type entries to schemas

petemoore:
Thanks!

### DIFF
--- a/schemas/artifact-created-message.yml
+++ b/schemas/artifact-created-message.yml
@@ -32,7 +32,7 @@ properties:
     title:        "Artifact Created"
     description: |
       Information about the artifact that was created
-    type:         array
+    type:         object
     properties:
       storageType:
         title:    "Artifact Storage-Type"
--- a/schemas/artifact-created-message.yml
+++ b/schemas/artifact-created-message.yml
@@ -24,7 +24,7 @@ properties:
     description: |
       Identifier for the worker within which the run with the created artifact
       is running.
-    type: "string"
+    type:         string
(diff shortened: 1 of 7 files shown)


schemas/artifact-created-message.yml
```diff
@@ -24,7 +24,7 @@ properties:
     description: |
       Identifier for the worker within which the run with the created artifact
       is running.
-    type: "string"
+    type:         string
```
petemoore: This is just to keep consistent formatting

schemas/artifact-created-message.yml
```diff
@@ -32,12 +32,14 @@ properties:
     title:        "Artifact Created"
     description: |
       Information about the artifact that was created
+    type:         array
```
jonasfj: type object if I'm not mistaken it has properties..

schemas/constants.js
```diff
@@ -79,6 +79,7 @@ module.exports = {
   // Message version numbers
   "message-version": {
     "description":  "Message version",
+    "type":         "string",
```
jonasfj: integer

schemas/artifact-created-message.yml
jonasfj: nice! :)

schemas/artifact-created-message.yml
petemoore: whoooooops

schemas/constants.js
petemoore: whoooooops again

PR #61, opened 2015-10-28
Issue temp creds

jonasfj:
The interesting bits are in `v1.js` and `artifacts.js` + schema changes...

Read the docs in `constants.js` to see what the temp creds are for... Also this includes the task definition in return value from `claimTask`

### DIFF
--- a/routes/v1.js
+++ b/routes/v1.js
@@ -993,7 +993,7 @@ api.declare({
     start:  new Date(Date.now() - 15 * 60 * 1000),
     expiry: new Date(takenUntil.getTime() + 15 * 60 * 1000),
     scopes: [
-      'queue:task-claim:' + taskId + '/' + runId
+      'queue:claim-task:' + taskId + '/' + runId
     ].concat(task.scopes),
     credentials: this.credentials
   });
@@ -1020,6 +1020,8 @@ api.declare({
     [
       'queue:claim-task',
       'assume:worker-id:<workerGroup>/<workerId>'
+    ], [
+      'queue:claim-task:<taskId>/<runId>'
     ]
   ],
   deferAuth:  true,
@@ -1054,8 +1056,10 @@ api.declare({
 
   // Authenticate request by providing parameters
   if (!req.satisfies({
+    taskId,
+    runId,
     workerGroup:    run.workerGroup,
-    workerId:       run.workerId
+    workerId:       run.workerId,
   })) {
     return;
   }
@@ -1113,7 +1117,7 @@ api.declare({
     start:  new Date(Date.now() - 15 * 60 * 1000),
     expiry: new Date(takenUntil.getTime() + 15 * 60 * 1000),
     scopes: [
-      'queue:task-claim:' + taskId + '/' + runId
+      'queue:claim-task:' + taskId + '/' + runId
     ].concat(task.scopes),
     credentials: this.credentials
   });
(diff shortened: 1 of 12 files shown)


jonasfj:
don't worry about travis test... it works locally... travis is inherently broken...

petemoore:
> don't worry about travis test... it works locally... travis is inherently broken...

I suspect this is a concurrency issue due to the PR job and the push job running in parallel, which we should probably fix.

routes/v1.js
```diff
@@ -986,13 +988,25 @@ api.declare({
     takenUntil:   run.takenUntil
   }, task.routes);
 
+  // Create temporary credentials for the task
+  let credentials = taskcluster.createTemporaryCredentials({
+    start:  new Date(Date.now() - 15 * 60 * 1000),
```
gregarndt: Is the 15 minutes to allow for some drift?

jonasfj: Yeah... Yeah, I see no reason to be super strict...
It annoying with race conditions :)

routes/v1.js
```diff
@@ -1040,8 +1056,10 @@ api.declare({
 
   // Authenticate request by providing parameters
   if (!req.satisfies({
+    taskId,
+    runId,
     workerGroup:    run.workerGroup,
-    workerId:       run.workerId
+    workerId:       run.workerId,
```
gregarndt: nit: is the trailing comma intentional? I didnt' see that style elsewhere.

jonasfj: Babel allows it... And when we decide to lint it'll probably be required...

So I'm just trying out the latest fashion :)

gregarndt: Ah, wait to stay on top of the latest trailing comma craze :)  I just didn't know because you added a couple of things to the end of an object (line 1009) and didn't have a trailing comma :)

schemas/task-reclaim-response.yml
```diff
@@ -0,0 +1,41 @@
...
+  workerId:
+    description: |
+      Identifier for the worker executing this run.
+    type:         string
+    minLength:    {$const: identifier-min-length}
+    maxLength:    {$const: identifier-max-length}
+    pattern:      {$const: identifier-pattern}
+  takenUntil:
+    description: |
+      Time at which the run expires and is resolved as `failed`,
```
gregarndt: Is it failed or marked as an exception with reason "claim-expired" ?

schemas/task-reclaim-response.yml
```diff
@@ -0,0 +1,41 @@
...
+    type:         string
+    format:       date-time
+  credentials:  {$const: task-credentials}
+additionalProperties: false
+required:
+  - status
+  - runId
+  - workerGroup
+  - workerId
+  - takenUntil
```
gregarndt: so "credentials" aren't required, but don't we always return them no matter what? I mean, it's not as if the caller has the option of not getting temp creds sent back is there?

jonasfj: Good catch, will fix!!! :)

schemas/task-reclaim-response.yml
jonasfj: Copy pasted from the old... But let's fix it now...

gregarndt:
:ship:

jonasfj:
will ship it... wow can't believe a test on travis passed...

PR #62, opened 2015-10-29
Bug 1219436: only require 'queue:cancel-task' to cancel a task

djmitche:


jonasfj:
Note sure about this yet...
Maybe `queue:cancel-task:<schedulerId>/<taskGroupId>/<taskId>` hmm... like I said not sure.

djmitche:
In that pattern, `<schedulerId>` is a singleton (`task-graph-scheduler`) while the other two are slugids and thus will always be `*` in a role (assuming we're not modifying roles for every task graph!).

I could see that being useful if we used temporary credentials -- so for example the scheduler returns a temp cred with `queue:cancel-task:task-graph-scheduler/$taskGraphId/*` from createTaskGraph; those creds can then be used to cancel (and, perhaps, extend?) the graph.

For the moment, though, i just want to get rid of the assume requirement.

djmitche:
This is getting subsumed by a larger project @jonasfj is working on

PR #63, opened 2015-11-03
Bug 1221045 - fix to schemas and routes

petemoore:
Two schema cleanups in this PR.

Cleanup 1: Task Definitions
===========================

Problem description
-------------------

We currently have a schema for validating task definition inputs to the queue, and a separate one for validating task definition outputs from the queue. The validation is essentially the same, except:

  a) For inputs, we provide defaults for values that can be omitted. Since the queue returns all fields, we don't support defaults for validating output from the queue.
  b) The property taskGroupId may be omitted as an input, and if the task-graph-scheduler is used, the taskGroupId will be set to the taskId. Therefore it is a "required" field as an output Task Definition from the queue, yet is not a required field as an input.

Reason to change
-----------------------

There are two minor cons of using two separate schemas, which is why I am looking to unite them:

  A) The two schemata can get out of sync for shared fields. E.g. we see already that retries has a maximum of 50 as an input, but 49 as an output. Also "Task Priority" has no type set for input schemas, but the type "string" for output schemas. These subtle differences are easily introduced when definitions exist in two places.

  B) In the generated code for taskcluster clients, we get duplicated code with ugly variable names, e.g. see

* https://github.com/taskcluster/taskcluster-client-go/blob/master/queue/queue.go#L539-L629
* https://github.com/taskcluster/taskcluster-client-go/blob/master/queue/queue.go#L990-L1080

Alternative approach
----------------------

If we don't want to use exactly the same schema for inputs and outputs, another strategy might be to separate out the commonality of the input and output schemas, and then have two separate schemas for the inputs and outputs that inherit from this common schema. However, this may be a little over-engineered. Open to thoughts on this.

A partial solution could be to change Titles to "Task Definition Request" and "Task Definition Response" at least to avoid that we get structs named "TaskDefinition" and "TaskDefinition1".

Other Considerations
-------------------

There is not a strong need to have strong validation on output from the Queue, since even if we declare taskGroupId as a field, but not list it as required, it is still legal for the Queue to output it, which it can consistently do.

Cleanup 2: Task Reclaim Response
================================

The title of http://schemas.taskcluster.net/queue/v1/task-reclaim-response.json says "Task Claim Response" whereas it should say "Task Reclaim Response". This is a [trivial fix](https://github.com/taskcluster/taskcluster-queue/pull/63/files#diff-fc32e69d0e7f533be359fecdccc3e690R2).


schemas/task.yml
```diff
@@ -187,7 +193,6 @@ required:
   - provisionerId
   - workerType
   - schedulerId
-  - taskGroupId
```
jonasfj: These changes are bad... because we don't get any validation that output is correct.
Suddenly we won't get an error serverside if we return the wrong thing and clients will have to handle the case where `taskGroupId` is missing.

Also `priority` isn't required in the createTask call... There is a big difference in what is required.

petemoore:
OK, if we really need differences between input and output task definitions, can we at least abstract the commonality and reuse it in both?

Maybe we can also change `title` to be `Task Definition Request` and `Task Definition Response`?

schemas/task.yml
petemoore: > Also `priority` isn't required in the createTask call... There is a big difference in what is required.

It isn't required, but there is a default set, so should be ok.

In any case, I think I'll try to split it out so we have different required fields for inputs and outputs, but share the same data structure...

petemoore: > In any case, I think I'll try to split it out so we have different required fields for inputs and outputs, but share the same data structure (due to reasons A and B listed in "Reason to change" section above).

Sadly this will add complexity for the code generators, since setting up the inheritance essentially requires using [allOf semantics](http://spacetelescope.github.io/understanding-json-schema/structuring.html). Although this is doable, it is a lot of extra work for code generators, so I think I'll leave it for now. :(

schemas/create-task-request.yml
```diff
@@ -68,6 +68,7 @@ properties:
       Priority of task, this defaults to `normal` and the scope
       `queue:task-priority:high` is required to define a task with `priority`
       set to `high`. Additional priority levels may be added later.
+    type:           string
```
jonasfj: wow, how do we keep missing these... thanks!

PR #64, opened 2015-11-03
Bug 1221132 - improve wording

petemoore:
The docs didn't consistently define which scopes are required by a task submitter due to entries in the task definition payload.

I'd like to introduce consistent wording so that it is simple to have an overview of which scopes you require for submitting any particular task definition.

### DIFF
--- a/routes/v1.js
+++ b/routes/v1.js
@@ -259,7 +259,7 @@ api.declare({
     "routing-key: `<route>`, then when the AMQP message about the task is",
     "published, the message will be CC'ed with the routing-key: ",
     "`route.<route>`. This is useful if you want another component to listen",
-    "for completed tasks you have posted."
+    "for completed tasks you have posted.",
     "",
     "**Important** Any scopes the task requires are also required for creating",
     "the task. Please see the Request Payload (Task Definition) for details."
--- a/routes/v1.js
+++ b/routes/v1.js
@@ -256,11 +256,13 @@ api.declare({
     "",
     "**Task specific routing-keys**, using the `task.routes` property you may",
     "define task specific routing-keys. If a task has a task specific ",
-    "routing-key: `<route>`, then the poster will be required to posses the",
-    "scope `queue:route:<route>`. And when the an AMQP message about the task",
-    "is published the message will be CC'ed with the routing-key: ",
+    "routing-key: `<route>`, then when the AMQP message about the task is",
+    "published, the message will be CC'ed with the routing-key: ",
     "`route.<route>`. This is useful if you want another component to listen",
     "for completed tasks you have posted."
+    "",
+    "**Important** Any scopes the task requires are also required for creating",
+    "the task. Please see the Request Payload (Task Definition) for details."
   ].join('\n')
 }, async function(req, res) {
   var taskId  = req.params.taskId;
@@ -384,7 +386,7 @@ api.declare({
 
 /** Define tasks */
 api.declare({
-  method:     'post',
+  method:     'put',
   route:      '/task/:taskId/define',
   name:       'defineTask',
   scopes:     [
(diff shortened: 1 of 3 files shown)


routes/v1.js
```diff
@@ -384,7 +386,7 @@ api.declare({
 
 /** Define tasks */
 api.declare({
-  method:     'post',
+  method:     'put',
```
petemoore: ... I've naughtily slipped this extra unrelated fix in ... sorry!
NB: further down in line 416 we already claim it is idempotent

routes/v1.js
```diff
@@ -256,11 +256,13 @@ api.declare({
     "",
     "**Task specific routing-keys**, using the `task.routes` property you may",
     "define task specific routing-keys. If a task has a task specific ",
-    "routing-key: `<route>`, then the poster will be required to posses the",
-    "scope `queue:route:<route>`. And when the an AMQP message about the task",
```
jonasfj: You lost details on the `queue:route:<route>` scope.

routes/v1.js
jonasfj: this is a breaking change...  We can't add this... Well, we would have to facilitate a gradual migration...
I don't think it's worth it.

Note: the argument for using `post`  is that the resource `/task/:taskId/define` will not be created... This creates the resource `/task/:taskId`, hence calling it `put` is a bit off... Note, here I just abuse `post` to be anything that doesn't feel REST enough :)

Neither of this is more REST, both violates the principals a bit, so I don't see the benefit of changing it... Considering the overhead of fixing it everywhere..

petemoore: OK I hadn't realised that the `POST` was to a different location to the `GET`.

I think we can still fix this.

At the moment `createTask` is doing this right, with a `PUT` against `/task/<taskId>`. I propose we backwardly-compatibly modify the `createTask` request payload schema, adding a new property `scheduled` as boolean, default: `true`. Then, we would deprecate `defineTask` API endpoint, until at some point we can remove it altogether. The new mechanism to create a task without scheduling it would be to call `createTask` endpoint, with `"scheduled": false`. Then it is using PUT, respecting the correct resource location, and is truly idempotent and REST-compliant.

What do you think?

routes/v1.js
petemoore: That was intentional - it was moved into the payload definition.

The reason for introducing this change was that some scope requirements were defined in the API documentation (e.g. `queue:route:<route>` as above) yet some scope requirements were defined in the schema definition (e.g. `queue:task-priority:high`). For all the properties in the TaskDefinition, several fields have scope requirements on the poster of the task definition - yet only `queue:route:<route>` scope requirement was documented in the API definition, the others were all in the schema definition.

In this change I moved all this scope requirement information into the documentation of the fields they relate to, and put an **Important** heading in the API documentation to say _read the schema docs_ to find out extra scopes you might require to post a task. I also moved the information into a structured form, so the scope requirements were not mentioned in free text form arbitrarily in the middle of a definition - instead in a section called "Task submitter required scopes" at the end of the property definition.

routes/v1.js
petemoore: ... scheduling a task would then logically be issuing a `PATCH` against `/task/<taskId>` with json object `{"scheduled": true}` ...

jonasfj: I don't like putting `state` into the task definition...
Also using `PATCH` on `/task/<taskId>` for scheduleTask... then what about cancelTask, rerunTask... Should we then extend our API reference format to support methods that specify a hardcoded payload and have the same URL pattern...

We are not truly REST-compliant... And this is a problem through-out all our APIs, we are somewhat REST-like, sorry... Not pure-rest.


-----
@petemoore, So I noticed how badly we violated REST principals a while ago. I saw no easy way to fix it... and after thinking about it I can't find any value from fixing this. Other than it would be most REST-like.

To do REST correctly we would however have to make self-describing resources and paths... Such that navigating to queue.taskcluster.net/v1/ lists all the paths and explains what they do..
And if the `accept` header is `text/html` we should render replies to HTML before returning them...

All in all I see no value in doing this right. It would be awesome to do right for purity, clarity and  I'm sure it would solve problems down the road. But the hassle of doing would be huge.

IMO, doing it right would require some serious abstractions so we can't do the wrong thing. No such API abstractions exists now... As far as I know... So I think we should file this **under some future**, just like we did with the idea of doing versioning of everything correctly. It would require abstractions and engineering to do it right, things we're not willing to invest in now.

Also I seriously doubt super purity has much value. Azure and AWS aren't always pure REST either (well, it depends for sure).

routes/v1.js
jonasfj: I don't really care... this seems nit to me... The added warning is nice though! :)

PR #66, opened 2015-11-13
Upgrade tc base

jonasfj:
Moving to new config and new component loader...


jonasfj:
once, landed this will also move us to node 4 :)

Note, we'll hold off rollout til monday or maybe this weekend... as there is likely to be config issues.

jonasfj:
FYI, if you have a hard time decoding the sanity of the loader thing... here is a dependency map:
https://img42.com/8Kq2i

jhford:
nice!!!  I love the use of the component loader!

I'm curious what's stopping us from using babel-compile/tc-rules here, but it's not going to stop this review

R+ from me.

PR #67, opened 2015-11-20
Bug 1228012: Deprecate all assume: scopes in API endpoints

jonasfj:
Got rid of all the `assume:` scopes...

Well, I left them in for keeping compatibility...

IMO, the big non-trivial thing we have to move here is:
`defineTask` and `createTask`, those are called by clients directly or through scheduler... We need those to move to the new scopes, and remove the legacy scopes.

Once that is done, the road is pretty open for API end-points to list tasks by `taskGroupId` and implementation of new big graph scheduler.

### DIFF
--- a/routes/artifacts.js
+++ b/routes/artifacts.js
@@ -18,7 +18,7 @@ api.declare({
       'assume:worker-id:<workerGroup>/<workerId>'
     ], [
       'queue:create-artifact:<name>',
-      'queue:claim-task:<taskId>/<runId>'
+      'queue:claim-run-id:<taskId>/<runId>'
     ]
   ],
   deferAuth:  true,
--- a/routes/artifacts.js
+++ b/routes/artifacts.js
@@ -17,8 +17,7 @@ api.declare({
       'queue:create-artifact:<name>',
       'assume:worker-id:<workerGroup>/<workerId>'
     ], [
-      'queue:create-artifact:<name>',
-      'queue:claim-run-id:<taskId>/<runId>'
+      'queue:create-artifact:<taskId>/<runId>'
     ]
   ],
   deferAuth:  true,
--- a/routes/v1.js
+++ b/routes/v1.js
@@ -237,7 +237,16 @@ api.declare({
   route:      '/task/:taskId',
   name:       'createTask',
   idempotent: true,
-  scopes:     [['queue:create-task:<provisionerId>/<workerType>']],
+  scopes:     [
+    [
+      // Legacy scope option to be removed
+      'queue:create-task:<provisionerId>/<workerType>',
+    ], [
+      'queue:define-task:<provisionerId>/<workerType>',
+      'queue:task-group-id:<schedulerId>/<taskGroupId>',
+      'queue:schedule-task:<schedulerId>/<taskGroupId>',
+    ],
+  ],
   deferAuth:  true,
   input:      'create-task-request.json#',
   output:     'task-status-response.json#',
@@ -273,11 +282,19 @@ api.declare({
     return 'queue:route:' + route;
   });
 
+  // Patch default values and validate timestamps
+  var detail = patchAndValidateTaskDef(taskId, taskDef);
+  if (detail) {
+    return res.status(detail.code).json(detail.json);
+  }
+
   // Authenticate request by providing parameters, and then validate that the
   // requester satisfies all the scopes assigned to the task
   if (!req.satisfies({
     provisionerId:  taskDef.provisionerId,
-    workerType:     taskDef.workerType
+    workerType:     taskDef.workerType,
+    schedulerId:    taskDef.schedulerId,
+    taskGroupId:    taskDef.taskGroupId,
   }) || !req.satisfies([taskDef.scopes])
      || !req.satisfies([routeScopes])) {
     return;
@@ -287,12 +304,6 @@ api.declare({
   if (taskDef.priority !== 'normal' &&
       !req.satisfies([['queue:task-priority:' + taskDef.priority]])) {
     return;
-  }
-
-  // Patch default values and validate timestamps
-  var detail = patchAndValidateTaskDef(taskId, taskDef);
-  if (detail) {
-    return res.status(detail.code).json(detail.json);
   }
 
   // Insert entry in deadline queue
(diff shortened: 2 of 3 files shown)


jonasfj:
@petemoore, just review this instead... it's the implemenation of the etherpad...

petemoore:
notes for self
==========
etherpad: https://public.etherpad-mozilla.org/p/jonasfj-tc-queue-scope-rethink-big-scheduler

routes/artifacts.js
```diff
@@ -18,7 +18,7 @@ api.declare({
       'assume:worker-id:<workerGroup>/<workerId>'
     ], [
       'queue:create-artifact:<name>',
-      'queue:claim-task:<taskId>/<runId>'
+      'queue:claim-run-id:<taskId>/<runId>'
```
petemoore: Better than reusing `queue:claim-run-id:<taskId>/<runId>` would be `queue:create-artifact-for:<taskId>/<runId>` - otherwise we assume that whichever component is allowed to claim a task is also allowed to create an artifact for it. Imagine a fictitious "Jenkins Bridge" component whereby a central component claimed tasks, but individual Jenkins slaves posted artifacts. The roles would indeed be separate.

routes/v1.js
```diff
@@ -614,8 +636,11 @@ api.declare({
   name:       'rerunTask',
   scopes:     [
     [
+      // Legacy scopes
       'queue:rerun-task',
       'assume:scheduler-id:<schedulerId>/<taskGroupId>'
+    ], [
+      'queue:rerun-task:<schedulerId>/<taskGroupId>',
```
petemoore: Better to use `queue:rerun-task:<schedulerId>/<taskGroupId>/<taskId>`. Note, `queue:rerun-task:<schedulerId>/<taskGroupId>/*` would provide existing semantics, but allow finer grained access control if required, when a component should only be allowed to rerun an individual task in a task graph, rather than an arbitrary one.

routes/v1.js
```diff
@@ -731,8 +756,11 @@ api.declare({
   name:       'cancelTask',
   scopes:     [
     [
+      // Legacy scopes
       'queue:cancel-task',
       'assume:scheduler-id:<schedulerId>/<taskGroupId>'
+    ], [
+      'queue:cancel-task:<schedulerId>/<taskGroupId>',
```
petemoore: Better to use `queue:cancel-task:<schedulerId>/<taskGroupId>/<taskId>`. Note, `queue:cancel-task:<schedulerId>/<taskGroupId>/*` would provide existing semantics, but allow finer grained access control if required, when a component should only be allowed to cancel an individual task in a task graph, rather than an arbitrary one. This could otherwise be otherwise open us up for an attack strategy whereby somebody gets access to cancel a "sister" task of a task graph they managed to get access to by virtue of having e.g. extended the task graph with an unrelated job.

routes/v1.js
```diff
@@ -1236,10 +1271,11 @@ api.declare({
   name:       'reportCompleted',
   scopes: [
     [
+      // Legacy
       'queue:resolve-task',
       'assume:worker-id:<workerGroup>/<workerId>'
     ], [
-      'queue:claim-task:<taskId>/<runId>'
+      'queue:claim-run-id:<taskId>/<runId>'
```
petemoore: There is an assumption here that the same component that claims a task should report the state. In the case of an "adapter" worker, such as the fictitious "Jenkins Bridge", it may be that one (central) service claims tasks, but another connected service reports state and uploads artifacts. Better than `queue:claim-run-id:<taskId>/<runId>` here would be `queue:resolve-task:<taskId>/<runId>`. Initially, the Queue could provide this scope in temporary credentials to all workers that claim tasks, but later if we needed to restrict this, we would have a mechanism to enable us to do so (for example, purely for illustration; we might enforce a mechanism such that the fictitious Jenkins worker might not receive the `queue:resolve-task:<taskId>/<runId>` scope in the temporary credentials given to it when claiming a task, and the slaves might instead make a separate call to gain temporary credentials with this scope, by virtue of some other scope they may posses). This is just a way of future-proofing. As I say, to keep things as they currently are, Queue can give `queue:resolve-task:<taskId>/<runId>` in response to a task claim for now.

routes/v1.js
```diff
@@ -1267,10 +1303,11 @@ api.declare({
   name:       'reportFailed',
   scopes: [
     [
+      // Legacy
       'queue:resolve-task',
       'assume:worker-id:<workerGroup>/<workerId>'
     ], [
-      'queue:claim-task:<taskId>/<runId>'
+      'queue:claim-run-id:<taskId>/<runId>'
```
petemoore: Same comment applies here as for `reportCompleted`.

routes/v1.js
```diff
@@ -1300,10 +1337,11 @@ api.declare({
   name:       'reportException',
   scopes: [
     [
+      // Legacy
       'queue:resolve-task',
       'assume:worker-id:<workerGroup>/<workerId>'
     ], [
-      'queue:claim-task:<taskId>/<runId>'
+      'queue:claim-run-id:<taskId>/<runId>'
```
petemoore: Same comment applies here as for `reportCompleted`.

routes/artifacts.js
jonasfj: Sorry, docs probably needs an update, but `queue:claim-run-id:<taskId>/<runId>` is granted to the temp credentials returned from `claimTask`.

Hence, whomever has `queue:claim-run-id:<taskId>/<runId>` for a specific taskId and runId, has the claim on that task/run... Hence, it's a worker and workers should always be allowed to upload artifacts, right?

routes/v1.js
jonasfj: Okay, I can support this! :)

routes/v1.js
jonasfj: I think we should optimize for the common flow... where workers have  `queue:claim-run-id:<taskId>/<runId>` from the temp creds issued by claimTask... And not complicate it... we can always add another flow/scope options...

routes/artifacts.js
jonasfj: @petemoore, okay can we agree this one isn't necessary... as we can just not give `queue:create-artifact:` ?

routes/v1.js
jonasfj: I can agree to add `queue:resolve-task:<taskId>/<runId>`

jonasfj:
@petemoore, have another look at let me know...

petemoore:
Looks good to me! :+1:

petemoore:
@jonasfj I'll let you hit the merge button!

jonasfj:
landed..

jonasfj:
I mean deployed... :)

PR #69, opened 2015-12-04
Bug 1230344: add 'superseded' as an exception reason

djmitche:


jonasfj:
@djmitche, you're welcome to push to a branch on `taskcluster/taskcluster-queue` I usually delete branches in the PR after landing, anyways... 

That way you get the credentials for tests to pass... Anyway, I'll land and probably roll this out later today... Maybe submit a counter proposal for docs to make it a tiny bit more generic...

PR #70, opened 2015-12-05
Trying to make superseded more generic

jonasfj:
I propose this documentation instead as slightly more generic...

As I see it the queue cannot dictate the definition of `superseded` and workers must supply their own semantics for this.. Ie. they must define what artifacts it uploads... And what it means to resolve a task as superseded...

@djmitche, I'm still not super happy about this. We're making a super-special case that all task consumers now have to deal with in some way or another.

---

Crazy idea... maybe we have a method `reportSuperseded(taskId, runId, {taskId: <supersededByTaskId>, runId: <supersededByRunId>})`. Which would then transplant the superseded run (`<supersededByRunId>`) to the task... (obviously requiring that `<supersededByRunId>` is resolved).
Meaning that if `<supersededByRunId>` failed, then the run added is also failed, if `<supersededByRunId>` was completed then the run added is also completed. And we create alias artifacts for everything.

This would require us to add a `supersededBy: {taskId: ..., runId: ...}` to the representation of a run in the task status structure. This way consumers that wanted to see the task a superseded would still be able to see this. But general task consumers that might not care about superseded-logic would still be able to consume the task.
I think this falls very close to the proposal where we upload the same artifacts and set the same resolution as the task we was superseded by (and doing that would certainly be easier).

It seems wrong to have a task reported as superseded and then it can't be consumed by consumers who are unaware of the superseded feature. Because superseded just means a newer or more-relevant result have been produced in a separate task.

I'm tempted to accept this solution because we need the coalescence feature. But I seriously fear we're building a major footgun here. As all future consumers of tasks will have to deal with "what does superseded mean?". One isn't bad, but reason from task exception is already hard to use given that I added `internal-error` and `resources-unavailable` both of which I don't like :)
(granted it may just be me being crazy here, most task consumers should probably only care about task state completed/failed/exception, and not concern themselves with the `reason` for why the task was resolved as exception).

### DIFF
--- a/schemas/task-exception-request.yml
+++ b/schemas/task-exception-request.yml
@@ -39,11 +39,23 @@ properties:
       error from which it couldn't recover. The queue will not retry runs
       resolved with this reason, but you are clearly signaling that this is a
       bug in the worker code.
-      **Report `superseded`** if the task was determined to have been
-      superseded by another task, and its results are no longer needed.  It is
-      convention in this case to create an artifact entitled
-      `public/superseded-by` containing the taskId of the task that superseded
-      this one.
+      **Report `superseded`** if a task can be determined to have been
+      superseded by another task, meaning its results aren't of interest, the
+      worker may report the task as _superseded_. The exact semantics of this
+      feature is worker-specific, upon encountering you should consult the
+      worker-specific documentation. Workers are free to use this exception as
+      they see fit, but they **must** document the behavior. It's strongly
+      recommended that the worker uploads a JSON artifact titled
+      `public/superseded-by.json` with details of why the task was superseded,
+      and a `taskId` property pointing to the task that superseded it.
+      Example use-case: Consider a worker that computes the next prime greater
+      than some number N given in `task.payload`, and then stores this prime in
+      a global database. If upon claiming a task the worker does a lookup in
+      the database and observes that the next time for N is already computed,
+      then it may be useful to just report the task as `superseded`.
+      In other cases a worker could also provide a facility by which
+      task-specific could report the task as `superseded`, for example using a
+      special exit code (not recommended, just illustrative).
 additionalProperties:   false
 required:
   - reason


djmitche:
I'm really not sure what you're suggesting -- a bunch of statements in the comment above are contradictory or have an unclear antecedent.

As things stand, there's very little new stuff for consumers to deal with -- for the most part, queue consumers treat the exception reason as arbitrary text, at most translating it to colors in the UI.

Being superseded is behaviorally equivalent to being cancelled, but I'd like treeherder to show it in a different color (perhaps strikethrough?).  Superseded tasks are not executed, and none of their dependent tasks should be executed.  They do not have artifacts because they were not executed.  The superseded and superseding tasks are not identical: the former is just not worth performing anymore.  So any proposal to alias artifacts or transplant runs between tasks is at best incorrect and at worst crazy.

As for the artifact convention: it's just a convention.  I could put that in the "conventions" section of the documentation but it seems most relevant here.  The intent is to leave enough breadcrumbs for something like mozci to manually retrigger such jobs (I'm planning to add a `public/supersedes` artifact to the superseding build, too, which would allow a service to automatically bisect failed coalesced tasks).  I see this as no more harmful a convention than the `public/task.log` artifact: if a task doesn't upload such a thing, the UI looks a little funny -- big deal.

I still see coalescing as a core feature to any CI system -- it goes by many names, but the good ones all support it one way or another.  So I'm not particularly sympathetic to concerns about supporting it as a cross-cutting concern in TaskCluster.  That said, I think we can define it a little more broadly than "the worker consulted a service and was told to do this task and mark these others as not done".  The prime-number example's not too realistic, though.

From a CI perspective, tasks for a branch might be superseded when that branch is merged, by the tests run on the merge commit (which might even be more thorough).  Or perhaps there's a nightly roll-up test which runs fast on "real" hardware, and when that succeeds all pending per-checkin tasks (that ordinarily run on much slower emulators) are marked as superseded.  The next morning developers looking for results from their pushes get an indication of what happened (with more detail than "got cancelled") and a link to the results (perhaps they want to look at test logs, even though the tests passed).

Outside the world of CI, consider a "big data" analysis system.  During the exploratory phases of an analysis, the user might create a collection of low-resolution tasks looking for patterns.  Once a pattern is evident, the user creates higher-resolution (thus, slower and more expensive) tasks, and at that point any pending low-resolution tasks can be marked as superseded.  Visualization tools watching those low-resolution tasks can then follow the link to instead display the high-resolution results.

Perhaps TC is being used to drive a mechanical turk system, so people at terminals are claiming and executing tasks.  Tasks are created in multiples of 5, so that any two worker's mistakes can be "outvoted" by other workers executing the same task.  But once 3 tasks are complete with the same result, the other two no longer need to be executed and can be marked as superseded by the third.  If the user wants more confidence (perhaps an ML algorithm indicates the result is unlikely) they can re-trigger the superseded tasks.

In short, superseding is a generic concept for any task-execution framework and is best modeled as an exception type without any attempt to hide the fact that the task was not executed.

PR #71, opened 2016-01-12
Lock scheduler

jonasfj:
very small changes... test pass so it should work...

I wanted to change scopes to require `scheduler-id`, but we since introduced some saner scopes...

So now I'm merging in the good things from this a branch, a bunch of small minor improvements...

### DIFF
--- a/routes/v1.js
+++ b/routes/v1.js
@@ -280,9 +280,7 @@ api.declare({
   }
 
   // Find scopes required for task specific routes
-  var routeScopes = taskDef.routes.map(function(route) {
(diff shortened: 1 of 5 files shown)


gregarndt:
Looks good, are we going to add trailing comma to our linter rules?

jonasfj:
> Looks good, are we going to add trailing comma to our linter rules?
He he, maybe we should... This was just because I removed an extra scope I was looking to add..
Be honestly yeah, we should probably start using trailing comma more places..

Note: I still don't have any linter on any of this...

PR #72, opened 2016-01-14
List by task group

jonasfj:
https://bugzilla.mozilla.org/show_bug.cgi?id=1137914


Procfile
```diff
@@ -3,4 +3,6 @@ claimReaper:      babel-node src/main claim-reaper
 deadlineReaper:   babel-node src/main deadline-reaper
 expireArtifacts:  babel-node src/main expire-artifacts
 expireTask:       babel-node src/main expire-tasks
+expireTask:       babel-node src/main expire-task-groups
```
gregarndt: Don't all processes need to have  unique names?

config.yml
```diff
@@ -25,6 +27,11 @@ defaults:
     # taskcluster.fromNow, notice this should be negative!
     taskExpirationDelay:          '- 30 minutes'
 
+    # Time extension of the expires property for task-group definitions.
+    # This is properties holding the schedulerId for a task-group, hence, we
```
gregarndt: nit: "These are properties"

config.yml
```diff
@@ -25,6 +27,11 @@ defaults:
...
+    # expire them later to reduce risk of abuse
```
gregarndt: What kind of abuse would there be?

queue/data.js
```diff
@@ -171,7 +171,7 @@ Task.prototype.hasClaim = function() {
  *
  * Returns a promise that all expired tasks have been deleted
  */
-Task.expire = async function(now) {
+Task.expire = async function(now = new Date()) {
```
gregarndt: Are there places where we call expire() without any date?  I didn't see it, so I think I'm overlooking the need of the default value here.

routes/v1.js
```diff
@@ -163,6 +181,58 @@ api.declare({
...
+    "List taskIds of all tasks sharing the same `taskGroupId`.",
+    "",
+    "As a task-group main contain an unbounded number of tasks, this end-point",
+    "may return a `continuationToken`. To continue listing tasks you must",
+    "`listTaskGroup` again with the `continuationToken` as the query-string",
+    "option `continuationToken`.",
+    "",
+    "By default this end-point will try to return up to 1000 members in one",
+    "request. But it **may return less**, even if more tasks are available.",
+    "It may also return a `continuationToken` even though there is no more",
```
gregarndt: nit: "there are no more"

routes/v1.js
```diff
@@ -163,6 +181,58 @@ api.declare({
...
+  description: [
+    "List taskIds of all tasks sharing the same `taskGroupId`.",
+    "",
+    "As a task-group main contain an unbounded number of tasks, this end-point",
+    "may return a `continuationToken`. To continue listing tasks you must",
+    "`listTaskGroup` again with the `continuationToken` as the query-string",
+    "option `continuationToken`.",
+    "",
+    "By default this end-point will try to return up to 1000 members in one",
+    "request. But it **may return less**, even if more tasks are available.",
```
gregarndt: Why would this return less than the 1000 members even if there are more to return?

routes/v1.js
```diff
@@ -163,6 +181,58 @@ api.declare({
...
+    "use the query-string option `limit` to return fewer.",
+  ].join('\n')
+}, async function(req, res) {
+  let taskGroupId   = req.params.taskGroupId;
+  let continuation  = req.query.continuationToken || null;
+  let limit         = parseInt(req.query.limit || 1000);
+
+  let data = await this.TaskGroupMember.query({
+    taskGroupId,
+    expires: base.Entity.op.greaterThanOrEqual(new Date()),
```
gregarndt: Is this putting a filter on "expires" that will only return records that have an expires greater than or equal to now?

routes/v1.js
```diff
@@ -231,6 +301,60 @@ var patchAndValidateTaskDef = function(taskId, taskDef) {
...
+  );
+  if (!taskGroup) {
+    taskGroup = await ctx.TaskGroup.create({
+      taskGroupId,
+      schedulerId:  taskDef.schedulerId,
+      expires:      taskGroupExpiration,
+    }).catch(err => {
+      // We only handle cases where the entity already exists
+      if (!err || err.code !== 'EntityAlreadyExists') {
+        throw err;
```
gregarndt: If there is no error, what happens here when throwing `undefined`?

routes/v1.js
```diff
@@ -231,6 +301,60 @@ var patchAndValidateTaskDef = function(taskId, taskDef) {
...
+      }
+      return ctx.TaskGroup.load({taskGroupId});
```
gregarndt: Is this because of a race between a request coming in to load it, and another request that's creating it?

gregarndt: Is there an issue with returning the result of TaskGroup.load() when elsewhere it seems like we're returning a boolean.

routes/v1.js
```diff
@@ -305,6 +429,11 @@ api.declare({
     return;
   }
 
+  // Ensure group membership is declared, and that schedulerId isn't conflicting
+  if (!await ensureTaskGroup(this, taskId, taskDef, res)) {
+    return;
```
gregarndt: How will this cause the request to fail?  By doing an empty return here, does it cause a standard error and message to be returned?

test/api/artifact_test.js
```diff
@@ -315,6 +315,66 @@ suite('Post artifacts', function() {
     });
   });
 
+  test("Check expire doesn't drop table", async () => {
```
gregarndt: Are these tests related to this bug or just something that snuck in? :)

test/api/expiretask_test.js
```diff
@@ -63,4 +63,33 @@ suite('Task Expiration (expire-tasks)', function() {
       assume(err.statusCode).equals(404);
     });
   });
+
+  test("expire won't drop table", async () => {
```
gregarndt: what table are you hoping isn't dropped here with expiring this task?

test/api/taskgroup_test.js
```diff
@@ -0,0 +1,257 @@
...
+    payload:          {},
+    metadata: {
+      name:           "Unit testing task",
+      description:    "Task created during unit tests",
+      owner:          'jonsafj@mozilla.com',
+      source:         'https://github.com/taskcluster/taskcluster-queue'
+    }
+  };
+
+  test("Create two tasks", async () => {
```
gregarndt: is this testing anything related to the task group options? Just seems that it's making sure two tasks could be created.

test/api/taskgroup_test.js
```diff
@@ -0,0 +1,257 @@
...
+
+    debug("### Creating taskB");
+    let taskIdB = slugid.v4();
+    await helper.queue.createTask(taskIdB, _.defaults({
+      taskGroupId,
+    }, taskDef));
+
+    let result = await helper.queue.listTaskGroup(taskGroupId);
+    assert(!result.continuationToken);
+    assert(_.includes(result.members, taskIdA));
```
gregarndt: fyi, es6/7 stuff adds Array.includes()

test/api/taskgroup_test.js
```diff
@@ -0,0 +1,257 @@
...
+    assert(_.includes(result.members, taskIdB));
```
gregarndt: any value in ensuring it *only* includes those two?

Procfile
jonasfj: good point thanks...

config.yml
jonasfj: Create a task with different schedulerId in a task-group that already exists... even if the tasks in the group and the group definition is racing to expire.

queue/data.js
jonasfj: yeah, I just added it for fun... But it's probably safer to force people to specify `now`, as we always want some delay to be involved...

routes/v1.js
jonasfj: Because it's one azure call... It could happen in theory.. In practice probably won't... When coding against the API, it's just something worth noting, it won't really complicate peoples code... I hope not... Or if so, they should restructure it.

routes/v1.js
jonasfj: Yeah, that's the intention... I think we should do the same for artifacts and tasks...

We delete things after expires.. .But let's make then invisible when `expires < now()`.

routes/v1.js
jonasfj: I think it treats `undefined` or `null` as any other value... You can do `throw "my string"` in javascript...
You shouldn't and I doubt we have code that do, but I wanted to avoid an error saying that `null doesn't have a property code`

routes/v1.js
jonasfj: We're not returning `TaskGroup.load()` because we're inside a `.catch(err => {` handler... So really if `TaskGroup.create(...)` fails, `.catch(err => {...` get called and when we return a value from in here that becomes the result of the promise from `TaskGroup.create(...).catch(err => {...` (it's promise chaining... pretty neat..

routes/v1.js
jonasfj: `res` is passed to `ensureTaskGroup` which will send an error when it returns `false`..

It's a bit of a hack, but I avoid typing the same error message twice...

test/api/artifact_test.js
jonasfj: Extra that snuck in.... Feel free to ignore... Basically, I realized when writing expiration tests for taskgroup thing, that our tests for expiration only tests that things gets deleted... They don't test that things that aren't suppose to expire doesn't get deleted...

If by accident we flipped  `expires <= now()` into `expires >= now()` then we would effectively drop the entire table :)
I think that would be bad... So these tests are just to guard against that...

test/api/expiretask_test.js
jonasfj: The task table...

The expires test we had, only tested that things got deleted... You can delete things by dropping the entire table, and our test would still pass... So this is a test to prove that expiration of tasks doesn't drop the entire table. He he :)

test/api/taskgroup_test.js
jonasfj: Two tasks with the same taskGroupId

test/api/taskgroup_test.js
jonasfj: yes, @jhford has been complaining that it doesn't work when compiling with babel... queue still runs under `babel-node` but I really ought to start compiling it...

test/api/taskgroup_test.js
jonasfj: Added `assert(result.members.length === 2);` good point, thanks...

gregarndt:
lgtm

routes/v1.js
```diff
@@ -163,6 +181,58 @@ api.declare({
...
+    continuationToken: /./,
+    limit: /^[0-9]+$/,
+  },
+  name:   'listTaskGroup',
+  output: 'list-task-group-response.json#',
+  title:  "List Task Group",
+  description: [
+    "List taskIds of all tasks sharing the same `taskGroupId`.",
+    "",
+    "As a task-group main contain an unbounded number of tasks, this end-point",
```
petemoore: main -> may

petemoore:
@jonasfj I've created https://github.com/taskcluster/taskcluster-base/pull/57 for updating the API reference schema to include query string - can you take a look, and merge if you are happy? This is currently breaking the go/java client auto-gen. Thanks!

routes/v1.js
jonasfj: Fixed and pushed now... outside this review... thanks :)

PR #73, opened 2016-01-15
bug 1237308

jonasfj:


routes/artifacts.js
```diff
@@ -519,6 +523,7 @@ api.declare({
   method:     'get',
   route:      '/task/:taskId/runs/:runId/artifacts',
   name:       'listArtifacts',
+  stability:  base.API.stability.experimental,
```
gregarndt: why is this experimental? (also the same goes for listLatestArtifacts)

routes/v1.js
```diff
@@ -766,6 +772,7 @@ api.declare({
   method:     'post',
   route:      '/task/:taskId/rerun',
   name:       'rerunTask',
+  stability:  base.API.stability.experimental,
```
gregarndt: This has been around for awhile and works...why is it experimental?

routes/artifacts.js
jonasfj: Basically, I'm not sure it's entirely stable yet...

For us to be able to iterate through all artifacts we likely have to support continuationToken...

routes/v1.js
jonasfj: I don't want people to call it... are you sure wish to keep it..

It sort of breaks the concept that a task only has one resolution... If rerunTask is called a task can have multiple pulse messages saying it's been resolved in different ways... That can't happen if rerunTask is removed...

Yes, we could still have multiple pulse messages, but they would just be duplicates... Not announcing different resolutions..

IMO, people shouldn't care about "runs".... But they do have to care about them now, because we can two runs that causes a resolution, if rerunTask is called.

gregarndt: multiple pulse messages are saying particular runs have been resolved in different ways...same way if we hit infra issues that cause a rerun.

gregarndt: I guess the difference is that there will only be one resolution of "completed" if we did not have the rerun endpoint.

jonasfj: infrastructure retries don't trigger pulse messages. I think that was the piece missing.

PR #74, opened 2016-01-22
Remove pending tasks scope

jonasfj:
For quick review...

### DIFF
--- a/routes/v1.js
+++ b/routes/v1.js
@@ -1634,10 +1634,13 @@ api.declare({
   output:     'pending-tasks-response.json#',
   title:      "Get Number of Pending Tasks",
   description: [
-    "Documented later...",
-    "This probably the end-point that will remain after rewriting to azure",
-    "queue storage...",
+    "Get an approximate number of pending tasks for the given `provisionerId`",
+    "and `workerType`.",
     "",
+    "The underlying Azure Storage Queues only promises to give us an estimate.",
+    "Furthermore, we cache the result in memory for 20 seconds. So consumers",
+    "should be no means expect this to be an accurate number.",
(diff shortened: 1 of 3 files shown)


routes/v1.js
```diff
@@ -1631,28 +1631,21 @@ api.declare({
...
   description: [
-    "Documented later...",
-    "This probably the end-point that will remain after rewriting to azure",
-    "queue storage...",
+    "Get an approximate number of pending tasks for the given `provisionerId`",
+    "and `workerType`.",
     "",
+    "The underlying Azure Storage Queues only promises to give us an estimate.",
+    "Furthermore, we cache the result in memory for 20 seconds. So consumers",
+    "should be no means expect this to be an accurate number.",
```
gregarndt: "by no means"

routes/v1.js
```diff
@@ -1631,28 +1631,21 @@ api.declare({
...
   output:     'pending-tasks-response.json#',
   title:      "Get Number of Pending Tasks",
   description: [
-    "Documented later...",
-    "This probably the end-point that will remain after rewriting to azure",
-    "queue storage...",
+    "Get an approximate number of pending tasks for the given `provisionerId`",
+    "and `workerType`.",
     "",
+    "The underlying Azure Storage Queues only promises to give us an estimate.",
```
gregarndt: Hrm, do we need to provide this level of details in our user facing documentation?  Seems like users don't care what Azure storage queue is or why this is just an estimate.

gregarndt:
lgtm

routes/v1.js
jonasfj: good point will drop that reference...

jonasfj:
oh, I forgot to merge this...

PR #75, opened 2016-02-08
Add first new reportError call

imbstack:
This is a first cut at replacing all of the plain ```res.status``` calls.


imbstack:
Oops, will do this again from a branch of the main repo.

PR #76, opened 2016-02-08
Update to new errors

imbstack:
This replaces our first ```res.status()``` with the new ```res.reportError()```. More will follow if this works.

### DIFF
--- a/test/api/reruntask_test.js
+++ b/test/api/reruntask_test.js
@@ -31,7 +31,6 @@ suite('Rerun task', function() {
     }
   };
 
-
   test("create, claim, complete and rerun (is idempotent)", async () => {
     var taskId = slugid.v4();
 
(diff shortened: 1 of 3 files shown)


test/api/reruntask_test.js
```diff
@@ -86,4 +86,11 @@ suite('Rerun task', function() {
     debug("### Requesting task rerun (again)");
     await helper.queue.rerunTask(taskId);
   });
-});
\ No newline at end of file
+
+  test("throw error on missing task", async () => {
+    var taskId = slugid.v4();
+    await helper.queue.rerunTask(taskId).catch( (err) => {
+        assert.equal(err.statusCode, 404);
```
jonasfj: We should also do `err.code == 'ResourceNotFound'`

test/api/reruntask_test.js
```diff
@@ -86,4 +86,11 @@ suite('Rerun task', function() {
     debug("### Requesting task rerun (again)");
     await helper.queue.rerunTask(taskId);
   });
-});
\ No newline at end of file
+
+  test("throw error on missing task", async () => {
+    var taskId = slugid.v4();
```
jonasfj: let's use `let`

jonasfj:
r+

PR #78, opened 2016-02-11
Update to all new-style errors in v1.js

imbstack:
The only places that the old-style is left is in artifacts.js
```
$ git grep res\.status ./routes/
routes/artifacts.js:    return res.status(400).json({
routes/artifacts.js:    return res.status(404).json({
routes/artifacts.js:    return res.status(404).json({
routes/artifacts.js:    return res.status(400).json({
routes/artifacts.js:      return res.status(409).json({
routes/artifacts.js:      return res.status(409).json({
routes/artifacts.js:      return res.status(409).json({
routes/artifacts.js:    return res.status(404).json({
routes/artifacts.js:    return res.status(403).json({
routes/artifacts.js:    return res.status(404).json({
routes/artifacts.js:    return res.status(404).json({
routes/artifacts.js:    return res.status(412).json({
routes/artifacts.js:    return res.status(404).json({
routes/artifacts.js:    return res.status(404).json({
routes/artifacts.js:    return res.status(412).json({
routes/v1.js:  res.status(200).json({

```

### DIFF
--- a/routes/v1.js
+++ b/routes/v1.js
@@ -503,8 +503,10 @@ api.declare({
     // offer an idempotent operation in that case
     if (!_.isEqual(taskDef, def) || task.runs.length === 0) {
       return res.reportError(
-        "RequestConflict",
-        "taskId {{taskId}} already used by another task.",
+        "RequestConflict", [
+          "taskId {{taskId}} already used by another task.",
+          "This could be the result of faulty idempotency!"
+        ].join('\n'),
         {
           taskId,
         });
@@ -754,7 +756,7 @@ api.declare({
   // Validate deadline
   if (task.deadline.getTime() < new Date().getTime()) {
     return res.reportError(
-      "ResourceExpired",
+      "RequestConflict",
       "Task {{taskId}} Can't be scheduled past it's deadline of {{deadline}}.",
       {
         taskId,
@@ -858,7 +860,7 @@ api.declare({
   // Validate deadline
   if (task.deadline.getTime() < new Date().getTime()) {
     return res.reportError(
-      "ResourceExpired",
+      "RequestConflict",
       "Task {{taskId}} Can't be rescheduled past it's deadline of {{deadline}}.",
       {
         taskId,
@@ -989,7 +991,7 @@ api.declare({
   // Validate deadline
   if (task.deadline.getTime() < new Date().getTime()) {
     return res.reportError(
-      "ResourceExpired",
+      "RequestConflict",
       "Task {{taskId}} Can't be cancelled past it's deadline of {{deadline}}.",
       {
         taskId,
@@ -1198,7 +1200,7 @@ api.declare({
       "Run {{runId}} not found on task {{taskId}}.",
       {
         taskId,
-        runId: runId.toString(),
+        runId
       });
   }
   // If the run wasn't claimed by this workerGroup/workerId, then we return
@@ -1211,7 +1213,7 @@ api.declare({
       "RequestConflict",
       "Run {{runId}} was already claimed by another worker.",
       {
-        runId: runId.toString(),
+        runId,
       });
   }
 
@@ -1301,7 +1303,7 @@ api.declare({
       "Run {{runId}} not found on task {{taskId}}.",
       {
         taskId,
-        runId: runId.toString(),
+        runId
       });
   }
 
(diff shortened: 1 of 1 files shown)


routes/v1.js
```diff
@@ -496,9 +502,12 @@ api.declare({
     // otherwise the task would have been created with defineTask, and we don't
     // offer an idempotent operation in that case
     if (!_.isEqual(taskDef, def) || task.runs.length === 0) {
-      return res.status(409).json({
-        message:      "taskId: " + taskId + " already used by another task"
-      });
+      return res.reportError(
+        "RequestConflict",
+        "taskId {{taskId}} already used by another task.",
```
jonasfj: This error message should probably state that it could be a faulty idempotency issues.

routes/v1.js
```diff
@@ -738,10 +753,13 @@ api.declare({
 
   // Validate deadline
   if (task.deadline.getTime() < new Date().getTime()) {
-    return res.status(409).json({
-      message:    "Task can't be scheduled past it's deadline",
-      error:      {deadline: task.deadline.toJSON()}
-    });
+    return res.reportError(
+      "ResourceExpired",
```
jonasfj: I think this is a conflict...  409, not a 410.

But also it's a conflict because you're trying to schedule something that already resolved, by deadline expiration.
The resources in question (the task) is still around, it still exists, it's just in a state where scheduling it isn't legal anymore.

routes/v1.js
```diff
@@ -839,10 +857,13 @@ api.declare({
...
```
jonasfj: Conflict again

routes/v1.js
```diff
@@ -961,10 +988,13 @@ api.declare({
 
   // Validate deadline
   if (task.deadline.getTime() < new Date().getTime()) {
-    return res.status(409).json({
-      message:    "Task can't be cancel task past it's deadline",
-      error:      {deadline: task.deadline.toJSON()}
-    });
+    return res.reportError(
```
jonasfj: Conflict, same argument as before

routes/v1.js
```diff
@@ -1270,9 +1317,13 @@ api.declare({
 
   // Check if task is past deadline
   if (task.deadline.getTime() <= Date.now()) {
-    return res.status(409).json({
-      message: "Task deadline exceeded"
-    });
+    return res.reportError(
+      "ResourceExpired",
```
jonasfj: Conflict again, I would say...
As the run we're trying to reclaim still exists.. .it's just not in a state where it can be reclaimed.

I think `410` is reserved for when something is "gone", as in the resource addressed isn't there anymore. We'll rarely have this, as we'll often just return 404, because the resource isn't there :) And we have no record that it ever was...

Maybe one day in artifacts, we'll store the record that the artifact existed after we delete the S3 object...  But we're not doing that now.. (ie. the artifact metadata record could exist until the task expires, but the s3 object could expire sooner -- in some future).

routes/v1.js
```diff
@@ -1311,9 +1362,13 @@ api.declare({
...
   if (task.runs.length - 1 !== runId || run.state !== 'running') {
-    return res.status(409).json({
-      message: "Run is resolved, or not running"
-    });
+    return res.reportError(
+      "RequestConflict",
+      "Run {{runId}} on task {{taskId}} is resolved or not running.",
+      {
+        taskId,
+        runId: runId.toString()
```
jonasfj: I've proposed: https://github.com/taskcluster/taskcluster-lib-api/pull/12

So we don't have to do `toString()` and ` task.deadline.toJSON()` :)
Let me know if you like it.

routes/v1.js
```diff
@@ -1353,17 +1408,24 @@ var resolveTask = async function(req, res, taskId, runId, target) {
...
 
   // Handle cases where the run doesn't exist
   var run = task.runs[runId];
   if (!run) {
-    return res.status(404).json({
-      message: "Run not found"
-    });
+    return res.reportError(
+      "ResourceNotFound",
+      "Run {{runId}} not found on task {{taskId}}.",
```
jonasfj: btw, I love the improvement of these error messages... By all means make them more explanatory :)

jonasfj:
Mostly nits, but we have a few places we use `ResourceExpired` incorrectly that's not good. In some cases it could break existing clients because they rely on the http statusCode..

(yeah, before this we had to resort to that... now we'll hopefully use .code)

jonasfj:
`r+`

PR #79, opened 2016-02-15
Initial work for transitioning to cloud-mirror

jhford:
@jonasfj : I would like to do short review cycles with you on these patches to
make sure this is done correctly.  Please let me know if using tags is a good
way to do this.  For creating jobs which would have the useCloudMirror tag, is
it just an extra entry into the tags dict when submitting the task/taskgraph?

### DIFF
--- a/routes/artifacts.js
+++ b/routes/artifacts.js
@@ -4,6 +4,7 @@ var _       = require('lodash');
 var assert  = require('assert');
 var Promise = require('promise');
 var base    = require('taskcluster-base');
+var urllib  = require('url');
 
 // Maximum number of artifacts to list
 const MAX_ARTIFACTS_LISTING = 10 * 1000;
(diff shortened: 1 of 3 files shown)


jhford:
this doesn't break tests, this is more to just verify that the approach is correct and that i'm putting code in the right place

routes/artifacts.js
```diff
@@ -361,13 +368,18 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
     var url = null;
-    var prefix = artifact.details.prefix;
-    if (artifact.details.bucket === this.publicBucket.bucket) {
-      var region = this.regionResolver.getRegion(req);
-      if (region) {
-        url = 'http://' + this.publicProxies[region] + '/' + prefix;
-      } else {
-        url = this.publicBucket.createGetUrl(prefix);
+    
+    if (task.tags.useCloudMirror) {
```
jonasfj: `task.extra`

jonasfj: yeah, task.extra is a good option...

----
For the record I fully support just moving over everything and once and see the world burn... It's much faster to revert if we see issues...

jonasfj: granted I'm crazy :)

jhford: hahaha.  i'm not a fan of burning everything!

jhford:
@jonasfj what do you think about this version of the patch?

routes/artifacts.js
```diff
@@ -344,11 +345,18 @@ api.declare({
...
+        taskId:     taskId,
+        runId:      runId,
+        name:       name
+      }, true),
+      this.Task.load({
+        taskId: taskId
+      }, true),
+  ]);
+  var artifact = promiseResponse[0];
+  var task = promiseResponse[1];
```
jonasfj: I don't like this at all...
The pattern should be:
```js
var [task, artifact] = await Promise.all([
  ...
]);
```
And even then I don't like it... Because we shouldn't load the task definition here... Why?

I know you are doing a temporary hack... So I'll allow it for now... Please label this with a comment `// TEMPORARY HACK: Remove this.Task.load(...) in the future!`

Otherwise, we might forget to remove `Task.load` when we clean up...

routes/artifacts.js
```diff
@@ -361,19 +369,74 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+
+      if (artifact.details.bucket === this.publicBucket.bucket) {
+        // The implementation of taskcluster at taskcluster.net doesn't
+        // currently use us-east-1 as our canonical datacenter, but we want to
+        // make sure that we could support it.  Like in the could mirror
+        // service, we have to account for us-east-1 behaving slightly
+        // differently
+        var vhost;
+        if (this.artifactRegion === 'us-east-1') {
+          vhost = 's3.amazonaws.com';
```
jonasfj: Wrong... But I guess it doesn't matter

src/main.js
```diff
@@ -237,6 +237,8 @@ let load = base.loader({
         regionResolver:   ctx.regionResolver,
         publicProxies:    ctx.cfg.app.publicArtifactBucketProxies,
         credentials:      ctx.cfg.taskcluster.credentials,
+        cloudMirrorHost:  ctx.cfg.app.cloudMirrorHost,
+        artifactRegion:   ctx.cfg.aws.region,
```
jonasfj: Please label this as a hack that will be removed!

At-least I strongly suspect we should...

jonasfj:
`r+` because it's a temporary hack...  :)

@jhford, before we move this to 100% on, let's move region resolution into cloud-mirror... I don't want to have to refactor that later...
Also please, consider improving cloudmirror, simplifying abstractions and reducing number of concepts. I've outlined a few suggestions a few times. Certainly having just one "address" concept is both feasible and realistic. Also getting rid of redirect support or dramatically simplifying the backend abstraction by moving it into a function should be reasonable.
I'm have very little confidence in my review, because I was unable to see how things fit together...

jonasfj:
@jhford, you'll want to monitor when you roll this out...

PR #80, opened 2016-02-23
Fix error message when reclaiming past deadline

gregarndt:


jonasfj:
lol, test are broken... not the fault of this PR...

`r+`, me (or @imbstack maybe) should be able to fix the test issue and land this... I think it's not using fakeAuth...

jonasfj:
Sorry for taking forever about landing this...

PR #81, opened 2016-03-01
task.dependencies

jonasfj:
Work in progress adding `task.dependencies` and `task.dependencyRelation`.

At this stage I need to write tests... try to actually run the code and fix all the bugs  that comes up :)

But it's also a decent time to start review for sanity...
I'm fairly confident I've designed it such that a crash anytime during `createTask`, doesn't prevent the same `createTask` call from succeeding if retried later... It's hard to get the idempotency correct, but I think it's sane... please have a look.


routes/v1.js
```diff
@@ -625,35 +658,45 @@ api.declare({
...
+  let deadline = new Date(taskDef.deadline);
+  let expires = new Date(taskDef.expires);
+
   // Insert entry in deadline queue (garbage entries are acceptable)
-  var deadline = new Date(taskDef.deadline);
   await this.queueService.putDeadlineMessage(taskId, deadline);
 
+  // Ensure we have a self-dependency, this is how defineTask works now
+  if (!_.includes(taskDef.dependencies, taskId)) {
+    taskDef.dependencies.push(taskId);
```
jonasfj: @imbstack, so funny/scary story, this created a problem because our current validator just sets the default value from the json-schema without making a clone.

Hence, by doing `dependencies.push(...` I actually modified the default value inside the validator.

imbstack:
I'm reading this today, I swear.

routes/v1.js
```diff
@@ -456,39 +465,49 @@ api.declare({
...
+  let deadline = new Date(taskDef.deadline);
+  let expires = new Date(taskDef.expires);
+
   // Insert entry in deadline queue
-  var deadline = new Date(taskDef.deadline);
   await this.queueService.putDeadlineMessage(taskId, deadline);
 
   // Try to create Task entity
   try {
+    let runs = [];
```
imbstack: A note for myself for later (this may be wrong): It looks like we always create tasks no matter whether or not they're ready to run.  `runs` will just be empty until they're ready to actually run.

routes/v1.js
```diff
@@ -456,39 +465,49 @@ api.declare({
...
+      tags:               taskDef.tags,
+      extra:              taskDef.extra,
+      runs:               runs,
+      takenUntil:         new Date(0),
     });
-  }
-  catch (err) {
+  } catch (err) {
     // We can handle cases where entity already exists, not that, we re-throw
     if (!err || err.code !== 'EntityAlreadyExists') {
```
imbstack: Why are we checking for `!err` here? Since this is inside `catch (err)` won't there always be an `err`?

routes/v1.js
```diff
@@ -502,52 +521,60 @@ api.declare({
...
-          "taskId {{taskId}} already used by another task.",
-          "This could be the result of faulty idempotency!"
-        ].join('\n'),
-        {
-          taskId,
-        });
+      return res.reportError('RequestConflict', [
+        "taskId {{taskId}} already used by another task.",
+        "This could be the result of faulty idempotency!",
+        "Existing task definition was: {{existingTask}}",
```
imbstack: This is going to be inserting a big json blob here, right? Maybe we can stick it on it's own line or think about moving in the direction of using Markdown here that we've talked before, right? We could stick this in a code block.

routes/v1.js
```diff
@@ -502,52 +521,60 @@ api.declare({
...
-      status:   status
-    });
+  // Construct task status, as we'll return this many times
+  let status = task.status();
+
+  // If first run isn't unscheduled or pending, all message must have been
+  // published before, this can happen if we came from the catch-branch
+  // (it's unlikely to happen). But no need to publish messages again
+  let runZeroState = task.runs[0].state;
+  if (runZeroState !== 'unscheduled' && runZeroState !== 'pending') {
```
imbstack: We do this all over the place and it's not an issue with this line or this PR in particular, but I keep getting an itch to use something like an enum in these situations to avoid typo bugs. Maybe just a hash of states?
```js
let states = {unscheduled: 'unscheduled', pending: 'pending'};
// ...
if (runZeroState !== states.unscheduled && runZeroState !== states.pending) {
// ...
}
```
And now the part where you explain to me why this is a dumb idea can happen!

routes/v1.js
```diff
@@ -502,52 +521,60 @@ api.declare({
...
-      // Publish task-defined message, we want this arriving before the
-      // task-pending message, so we have to await publication here
-      await this.publisher.taskDefined({
-        status:         status
-      }, task.routes);
+  // If first run is pending we publish messages about this
+  if (runZeroState === 'pending') {
+    await Promise.all([
+      // Put message into the task pending queue
+      this.queueService.putPendingMessage(task, 0),
```
imbstack: Just reading this as a novice to this codebase, what's the difference between this line and the subsequent one ([line 572](https://github.com/taskcluster/taskcluster-queue/pull/81/files#diff-ebe0840faded7b2e7bc8c6e798d1b159R572) at the moment).

routes/v1.js
```diff
@@ -1040,18 +1063,19 @@ api.declare({
   // Construct status object
   var status = task.status();
 
-  // Publish message about last run, if it was canceled
+  // If the last run was canceled, resolve dependencies and publish message
   if (run.state === 'exception' && run.reasonResolved === 'canceled') {
```
imbstack: `canceled` is a good reason to think about enum, imho. [I always want to spell it with two l's](http://grammarist.com/spelling/cancel/).

schemas/create-task-request.yml
```diff
@@ -44,6 +44,34 @@ properties:
...
+    description: |
+      The tasks relation to its dependencies. This property specifies the
+      semantics of the `task.dependencies` property.
+      If `on-completed` is given the task will be scheduled when all
+      dependencies are resolved _completed_ (successful resolution).
+      If `on-resolved` is given the task will be scheduled when all dependencies
+      have been resolved, regardless of what their resolution is.
+    type:           string
+    enum:           {$const: dependency-relation}
+    default:        'on-completed'
```
imbstack: Do we really need the `on-` part of these? Maybe just `completed` and `resolved`. I don't really care either way, it just seems more straightforward that way.

imbstack:
I'm about 3/4 way through reviewing this. Just thought that it might be a good idea to get the sentry-catches-crashes thing in here before this is shipped and rebase this on top of that change. It could save us a good bit of sadness down the road. What do you think?

queue/data.js
```diff
@@ -1,5 +1,5 @@
 var base    = require('taskcluster-base');
-var debug   = require('debug')('queue:data');
+var debug   = require('debug')('app');
```
imbstack: :+1:

queue/dependencytracker.js
```diff
@@ -0,0 +1,279 @@
...
+      let msg = '';
+      if (missing.length > 0) {
+        msg += '`task.dependencies` references non-existing tasks: \n';
+        msg += missing.map(taskId => {
+          return ' * ' + taskId + ','
+        }).join('\n') + '\n';
+        msg += 'All taskIds in `task.dependencies` **must** exist\n';
+        msg += 'before the task is created.\n';
+      }
+      if (expiring.length > 0) {
```
imbstack: We make this assertion so that we're free to schedule a task all the way up to its deadline, right? Without this assertion we'll sometimes still be able to succeed, but there would be situations where things don't work anymore, right?

queue/dependencytracker.js
```diff
@@ -0,0 +1,279 @@
...
+    await Promise.all(task.dependencies.map(async (requiredTaskId) => {
+      let requiredTask = await this.Task.load({taskId: requiredTaskId}, true);
+
+      // If task is missing, we should report and error
+      if (!requiredTask) {
+        return missing.push(requiredTaskId);
+      }
+
+      // Check if requiredTask expires before the deadline
+      if (task.deadline.getTime() > requiredTask.expires.getTime()) {
```
imbstack: Is this inequality backwards? I'm not really sure.

imbstack:
Ok, I've read this. I want to spend some more time with it, but it seems pretty good.

routes/v1.js
jonasfj: it's just an old leftover in case we somewhere did `throw null`... it shouldn't happen, but with promises sometime a promise can be rejected without an error...

I don't think we have to do this everywhere.. I think I was just being very defensive when I wrote this :)

routes/v1.js
jonasfj: he he, correctness is enforced by JSON schema for input and output...

Having an enum-object-like-thing like `reasons.CANCELLED` whose value is the string  would look nice.
But if we referenced a property that doesn't exist we would get `undefined` which is a value different from `run.reasonResolved`.

So I'm basically not sure how to do enums better... At least this way it's clear that it's fragile...

routes/v1.js
jonasfj: `queueService` is azure queue... `publisher` is pulse.

queue/dependencytracker.js
jonasfj: Yeah, and if one of your dependencies expires before the task... someone could technically reuse that taskId... And resolved it completed... using that as an attackvector to trick a task into running...

See comment I added below this... in source code...

queue/dependencytracker.js
jonasfj: I think it's right... if deadline of the current task is greater than expiration of a dependency something is bad...
So we add the dependency (requiredTask) to `expiring` used for reporting error.

jonasfj:
@imbstack, I think this is now done with tests and everything...

jonasfj:
@jhford you are also invited to have a look around :)

queue/dependencyresolver.js
```diff
@@ -0,0 +1,101 @@
...
+   * {
+   *   dependencyTracker:   // DependencyTracker instance
+   *   queueService:        // QueueService instance
+   *   pollingDelay:        // Number of ms to sleep between polling
+   *   parallelism:         // Number of polling loops to run in parallel
+   *                        // Each handles up to 32 messages in parallel
+   * }
+   */
+  constructor(options = {}) {
+    assert(options,                   'options are required');
```
imbstack: :laughing:

queue/dependencyresolver.js
```diff
@@ -0,0 +1,101 @@
...
+      debug("Fetched %s messages", messages.length);
+
+      await Promise.all(messages.map(async (m) => {
+        // Don't let a single task error break the loop, it'll be retried later
+        // as we don't remove message unless they are handled
+        try {
+          await this.dependencyTracker.resolveTask(m.taskId, m.resolution);
+          await m.remove();
+        } catch (err) {
+          debug("[alert-operator] Failed to handle message: %j" +
```
imbstack: I wish we had logging levels.

src/main.js
```diff
@@ -1,22 +1,23 @@
...
-let BlobStore         = require('../queue/blobstore');
-let data              = require('../queue/data');
-let Bucket            = require('../queue/bucket');
-let QueueService      = require('../queue/queueservice');
-let raven             = require('raven');
-let EC2RegionResolver = require('../queue/ec2regionresolver');
-let DeadlineResolver  = require('../queue/deadlineresolver');
-let ClaimResolver     = require('../queue/claimresolver');
-let DependencyTracker = require('../queue/dependencytracker');
+let debug               = require('debug')('queue:main');
```
imbstack: I wish github was better about whitespace-only diffs.

test/api/dependency_test.js
```diff
@@ -0,0 +1,359 @@
...
+    deadline:         taskcluster.fromNowJSON('3 days'),
+    payload:          {},
+    metadata: {
+      name:           "Unit testing task",
+      description:    "Task created during unit tests",
+      owner:          'jonsafj@mozilla.com',
+      source:         'https://github.com/taskcluster/taskcluster-queue'
+    }
+  };
+/*
```
imbstack: What's with this humongous block of commented out code?

imbstack: Oh, duh. This was taken care of in a later commit.

imbstack:
I think I'd still like to get the sentry exception handling stuff in here before this ships, but that's not necessary.

r+ looks great to me.

routes/v1.js
jhford: spitballing, you could do a second level property resolution
a = { CANCELLED: {v: 'canceled'}}
then in your code, you could do
a.CANCELLED.v
and it would throw if you were to do
a.INVALIDENUM.v.

Ugly though...

schemas/constants.js
```diff
@@ -135,4 +135,13 @@ module.exports = {
 
   // Pattern for scope names, for when-ever that is useful
   "scope-pattern":   "^[\\x20-\\x7e]*$",
+
+  // Maximum number of dependencies a single task can have
+  "max-task-dependencies": 100,
```
jhford: why did we pick 100?

jhford:
Did a once-over of this and it looks good.  I'm a little curious about why we have the limit of dependent tasks at 100, but i get that we do want some sort of limit.

schemas/constants.js
jonasfj: It was a prettt round number :)

Obviously there is some limit, I don't know what it is... But 100 should serve most use-cases and force intermediate dummy tasks in the edge cases...

queue/dependencyresolver.js
jonasfj: Or reporting to sentry... Which seems like the better option here...

routes/v1.js
jonasfj: I considered making a.CANCELLED()
but if you forget to so the call or the .v thing we back to square one...

To get static safety for this I think one would have to write a babeljs plugin that can do something magic..

But short if crazy over engineered solutions, I think the best option is to ensure tests coverage... That way we get tests too :)

PR #82, opened 2016-03-16
fix cloud-mirror opt-in

jhford:


routes/artifacts.js
```diff
@@ -402,6 +403,11 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
           // are in the canonical region
           url = canonicalArtifactUrl;
         } else {
+          // If the region is not resolved to anything, we assume that it's in the canonical
+          // region.  This should go away when we move region resolving into cloud-mirror
+          if (!region) {
+            region = this.artifactRegion;
```
jonasfj: Should you make the condition above:
`if (!region || this.artifactRegion === region) {` ?


hmm... that's also wrong too, as if region isn't resolved we should return the cloudfront url...

If we don't have a region shouldn't we be using `this.publicBucket.createGetUrl(prefix);` as that is the url to the CDN...

routes/artifacts.js
```diff
@@ -386,18 +387,18 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+          host: bucket + '.' + vhost,
           pathname: prefix,
         });
         console.log('CLOUDMIRROR canonical artifact url: ' + canonicalArtifactUrl);
 
-        if (this.artifactRegion === region) {
+        if (!region || this.artifactRegion === region) {
           // Since we don't need a redirect, we just set things to what they
           // are in the canonical region
           url = canonicalArtifactUrl;
```
jonasfj: Still wrong...

We don't get CDN, we need to use `this.publicBucket.createGetUrl(prefix);`

routes/artifacts.js
```diff
@@ -386,18 +387,18 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
-          vhost = `s3-${this.canonicalArtifactRegion}.amazonaws.com`;
+          vhost = `s3-${this.artifactRegion}.amazonaws.com`;
         }
 
         // Generate the public URL for this artifact
         var canonicalArtifactUrl = urllib.format({
           protocol: 'https:',
-          host: vhost,
+          host: bucket + '.' + vhost,
           pathname: prefix,
```
jonasfj: All of this logic actually exists in `this.publicBucket.createGetUrl(prefix);`

I think we're better off using that... and then giving it an optional flag, like:
`this.publicBucket.createGetUrl(prefix, forceS3);`
So that if `forceS3` is true, then we don't use CDN (even if available).

Obviously, we default to `forceS3 = false`

routes/artifacts.js
jhford: ok!  i can make it do that.

routes/artifacts.js
```diff
@@ -386,22 +387,15 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
         if (this.artifactRegion === 'us-east-1') {
           vhost = 's3.amazonaws.com';
         } else {
-          vhost = `s3-${this.canonicalArtifactRegion}.amazonaws.com`;
+          vhost = `s3-${this.artifactRegion}.amazonaws.com`;
```
jonasfj: We don't need vhost for anything deadcode

jonasfj:
This is much better... :)

Remove the last dead code and ship it... `r+`.

PR #84, opened 2016-03-23
Employ continuationToken for listArtifact

jonasfj:
I don't think this breaks anything....
Query string parameters are optional, so old client will just keep on working as they do now :)
Notice that I didn't have to modify the old tests, just add a new test case.

right now we list up to 10k without continuation token... After this we'll only be able to list up to 1k without continuation token... But with continuation token we'll be unbounded...

I doubt anyone has more than 1k artifacts anyways... So this should do fine :)

### DIFF
--- a/routes/artifacts.js
+++ b/routes/artifacts.js
@@ -594,10 +594,34 @@ api.declare({
   let limit         = parseInt(req.query.limit || 1000);
   // TODO: Add support querying using prefix
 
-  let artifacts = await this.Artifact.query({
-    taskId: taskId,
-    runId:  runId,
-  }, {continuation, limit});
+  let [
+    task,
+    artifacts,
+  ] = await Promise.all([
+    this.Task.load({taskId}, true),
+    this.Artifact.query({taskId, runId}, {continuation, limit})
+  ]);
+
+  // Give a 404 if not found
+  if (!task) {
+    return res.reportError(
+      'ResourceNotFound',
+      "No task with taskId: {{taskId}} found",
+      {taskId}
+    );
+  }
+
+  // Check that we have the run
+  if (!task.runs[runId]) {
+    return res.reportError(
+      'ResourceNotFound',
+      "Task with taskId: {{taskId}} run with runId: {{runId}}\n" +
+      "task status: {{status}}", {
+      taskId, runId,
+      status: task.status(),
+    });
+  }
+
 
   let result = {
     artifacts: artifacts.entries.map(artifact => artifact.json()),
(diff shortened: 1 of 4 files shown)


jonasfj:
This will fix: https://bugzilla.mozilla.org/show_bug.cgi?id=1240189

routes/artifacts.js
```diff
@@ -571,44 +568,45 @@ api.declare({
...
+    "you may limit this with the query-string parameter `limit`.",
   ].join('\n')
 }, async function(req ,res) {
-  var taskId = req.params.taskId;
-  var runId  = parseInt(req.params.runId);
+  let taskId        = req.params.taskId;
+  let runId         = parseInt(req.params.runId);
+  let continuation  = req.query.continuationToken || null;
+  let limit         = parseInt(req.query.limit || 1000);
   // TODO: Add support querying using prefix
```
gregarndt: Not something that changed in this PR, but just something I noticed is that we do no error checking here to ensure the task ID and run ID exist, but we do for the "/task/:taskId/artifacts" endpoint.

schemas/list-artifacts-response.yml
```diff
@@ -52,6 +52,17 @@ properties:
...
+  continuationToken:
+    type:             string
+    title:            "Continuation Token"
+    description: |
+      Opaque `continuationToken` to be given as query-string option to get the
+      next set of artifacts.
+      This property is only present if another request is necessary to fetch all
+      results. In practice the next request with a `continuationToken` may not
+      return additional results, but it can. Thus, you can only be sure to have
+      all the results if you've callwith `continuationToken` until you get a
```
gregarndt: "callwith"?

test/api/artifact_test.js
```diff
@@ -763,4 +763,77 @@ suite('Artifacts', function() {
...
+    assert(r2.continuationToken, "missing continuationToken");
+
+    debug("### listArtifacts, w. continuationToken");
+    let r3 = await helper.queue.listArtifacts(taskId, 0, {
+      continuationToken: r2.continuationToken,
+    });
+    assume(r3.artifacts.length).equals(1);
+    assume(r3.artifacts[0].contentType).equals('application/json');
+    assert(!r3.continuationToken, "unexpected continuationToken");
+    assume(r3.artifacts[0].name).not.equals(r2.artifacts[0].name);
```
gregarndt: Nice, I like that we check this!

gregarndt:
:+1:

routes/artifacts.js
jonasfj: @gregarndt, I guess we could check it... just because we want to return 404 rather than an empty list, right?

gregarndt: I was just mainly curious why we were checking it elsewhere but not here.  Nothing changed in this PR in regards to that so if we wanted to leave it unfixed or fix it in another PR at another time (just enter a bug?) then that's cool too.

jonasfj: just fixed it :)

PR #87, opened 2016-03-23
Babel compile and new validator

jonasfj:
Moving to babel-compile and node 5.x

Do you see anything particularly bad?


test/runtests.sh
```diff
@@ -2,27 +2,27 @@
 # USAGE: Run this file using `npm test` (must run from repository root)
```
imbstack: While we're at it, can we get rid of this and replace it in `package.json` with `test: mocha .test/*_test.js`?

imbstack:
r+

Hooray for spring cleaning!

imbstack:
Oh wait, isn't this removing `npm-shrinkwrap.json`? We don't really want to do that, right?

jonasfj:
Yeah, I'm upgrading modules.. so shrinkwrap will work again..

jonasfj:
@imbstack,
I just pilled on... this PR now includes new validator... Have look...

imbstack:
This looks good to me! r+

PR #88, opened 2016-04-04
[DONE] claimWork (ready to merge)

jonasfj:
Can I get a 30% review on this?

I'm not quiet sure if a single loop polling the queues is enough... Basically, I'm wondering if the through-put of this thing is high enough... or if it's arbitrarily limited by the linear nature of the polling loop...

@jhford, This is just for discussion... I'm not really sure I have to time finish this, I just wanted to open for debate on how to do and solve related problems...


src/api.js
```diff
@@ -1164,6 +1164,60 @@ api.declare({
...
+  let timer = null;
+  let aborted = new Promise(accept => {
+    timer = setTimeout(accept, 20 * 1000);
+    res.once('close', accept);
+  });
+
+  let result = await this.WorkClaimer.claim(
+    provisionerId, workerType, count, aborted,
+  );
+  clearTimeout(timer); // Just free up a timer no longer needed
```
jhford: i believe we'd also leak a bunch, maybe worth putting that in the comment :)

src/workclaimer.js
```diff
@@ -0,0 +1,140 @@
...
+
+/**
+ * HintPoller polls for hints for pending tasks.
+ *
+ * The azure queues don't know if a task is pending they just store hints of
+ * pending tasks. To be understood this way:
+ *  A) If a task is pending, there is a hint of the task in an azure queue,
+ *  B) If there is an hint in an azure queue, it may or may not be pending.
+ *
+ * It's an if, but not an only-if (think over-approximation).
```
jhford: I like this description.  it's consice and clear.  Should we also describe the case where there's a pending task but it does not have a hint -- or do we just consider that a dropped task?  This will be less relevant once workers can no longer access the queues directly.

src/workclaimer.js
```diff
@@ -0,0 +1,140 @@
...
+    while ((count = _.sumBy(this.requests, 'count')) > 0) {
+
+      let queues = this.owner.queueService.pendingQueues(
+        provisionerId, workerType,
+      );
+      for (let queue of queues) {
+        let hints = [];
+        while (count > 0 && (hints = await queue.poll(count)).length > 0) {
+          count -= hints.length;
+          this.claimHint
```
jhford: are you missing something here?  I feel like there's a missing argument list to a function and a missing implementation

jhford:
This looks good!  I don't know if there's a way to answer the through-put question other than load testing?

src/api.js
jonasfj: I think at some point timers elapse and get garbage collected, right? (so not really leaks)

jhford: maybe it's a gecko bug, but I know that leaking timers is a real concern in js for long lasting things

src/workclaimer.js
jonasfj: > Should we also describe the case where there's a pending task but it does not have a hint -- or do we just consider that a dropped task?

A pending task without a hint cannot have been successfully created.
`queue.createTask` won't return 200 unless the pending hint was en-queued. And if it fails with a random error or crash, the client should retry until it gets 200, or client will know for sure that the task wasn't successfully created.

Note: tasks that aren't successfully created, are typically resolved by deadline expiration...
But we don't really make any promises... I think that's fair.

src/workclaimer.js
jonasfj: yep

src/api.js
jonasfj: That's probably with `setInterval` or loops where you have the call back given to `setTimeout` calling `setTimeout`...
Otherwise, I would be very surprised.

jonasfj:
@imbstack, @jhford, please have a look through this..
I'll start working on a load-testing setup tomorrow, when I have an idea what the task through-put per second looks like I'll be ready to merge. But load-testing doesn't block deployment.

Note: I'm hoping for something like 30 tasks / second, but I wouldn't be surprised if it can't pull that off.
It's not super critical I suspect 10 tasks / second will be fine as run something like 10-15 dynos.

src/api.js
```diff
@@ -1282,6 +1284,59 @@ api.declare({
...
+  if (!req.satisfies({
+    workerGroup,
+    workerId,
+    provisionerId,
+    workerType,
+  })) {
+    return;
+  }
+
+  // Allow request to hand
```
imbstack: I'm not sure what this comment means

src/queueservice.js
```diff
@@ -634,6 +643,46 @@ class QueueService {
     return {queues, expiry};
   }
 
+  /**
+   * Return pending queues as list of poll(count) in.order of priority.
```
imbstack: Stray "." between in and order?

src/queueservice.js
```diff
@@ -634,6 +643,46 @@ class QueueService {
...
+   *   remove:  function() {} // Async function to delete the message
+   *   release: function() {} // Async function that makes the message visible
+   * }
+   */
+  async pendingQueues(provisionerId, workerType) {
+    // Find names of azure queues
+    let queueNames = await this.ensurePendingQueue(
+      provisionerId, workerType,
+    );
+    // Order by priorty (and convert to array)
```
imbstack: typo in priorty :p

src/workclaimer.js
```diff
@@ -0,0 +1,298 @@
...
+
+          // While we have hints and requests for hints we resolve requests
+          while (hints.length > 0 && this.requests.length > 0) {
+            let {resolve, count} = this.requests.shift();
+            resolve(hints.splice(0, count));
+          }
+
+          // Release remaining hints (this shouldn't happen often!)
+          await Promise.all(hints.map(hint => hint.release()));
+          this.parent._monitor.count('hints-released', hints.length);
```
imbstack: I love seeing this after a "this shouldn't happen often" comment! 💯

imbstack:
This is looking good. I owe it another good pass of review before I r+ it, but am happy so far.

src/api.js
jhford: I assume the next letters should be `le`, followed by what needs to be handled.

src/main.js
```diff
@@ -218,6 +219,19 @@ let load = base.loader({
...
   },
 
+  // Create workClaimer
+  workClaimer: {
+    requires: ['cfg', 'publisher', 'Task', 'queueService', 'monitor'],
+    setup: ({cfg, publisher, Task, queueService, monitor}) => new WorkClaimer({
+      publisher,
+      Task,
+      queueService,
+      monitor:        monitor.prefix('work-claimer'),
```
jhford: with `{propandval}` syntax, your indenting looks even wackier :)

src/workclaimer.js
```diff
@@ -0,0 +1,298 @@
...
+    // Get poll functions for pending queues (ordered by priority)
+    let polls = await this.parent._queueService.pendingQueues(
+      this.provisionerId, this.workerType,
+    );
+    // While we have requests for hints
+    while (_.sumBy(this.requests, 'count') > 0) {
+      let claimed = 0; // Count hints claimed
+
+      // In-order of priority, we poll hints from queues
+      for (let poll of polls) {
```
jhford: We had a discussion about how we should prioritize things with Jonas, Brian and myself.  In that discussion we agreed that there should be some sort of priority system better than all of the top priority then all of the next priority... stuff.  I think that's what's happening here.

If we have priority levels 1-5, 1 being highest priority, and we have 100 lvl2 jobs and that number isn't going down, any level3-5 jobs literally will never run in any case.  Maybe we should have some sort of fairness algorithm.  What about a round-robin where we take 10 of level 1, then 4 of level 2, then 3 of level 3, then 2 of level 4 and 1 of level 5.

Though, that said, for dealing with firedrills, we should probably consider the top priority as exclusive if it exists.

src/workclaimer.js
jhford: how not often... report to sentry not often?

Does sentry have a system to alert if you hit an error more than X times per unit time?  Can you report positives so that you can measure percentage failure and report on that?

jonasfj: it's not an error, it's a corner case... expected behavior if a connection times out while we're polling the queue we release the message back into the queue... That shouldn't happen often, so we measure how much it does happen.

src/main.js
jonasfj: true

src/workclaimer.js
jonasfj: > we agreed that there should be some sort of priority system better than all of the top priority then all of the next priority... stuff. I think that's what's happening here.

Next time you need to get that in writing :) hehe -- I recall the discussion I don't recall agreeing that we need nice-levels as oppose to strict priority levels. I had a discussion with catlee where he told me the current bb setup is strict priority.

Regardless, I maintain that if we want to shuffle the queues randomly every once in a while we can do so... But I want to see it being a requested feature.

TL;DR: We can do this later if we really want to. I don't think we want nice levels. But I'm not blocking in this on that discussion.

jonasfj:
@imbstack, I think this is ready to land, results of my load tests are:

```
createTask/claimTask/reportCompleted
  Running node lib/main load-test on ⬢ tc-queue-load-test... up, run.9665
  18.21 req/s success: 3277, failed: 0   // 16 loops in parallel for 3 min
  27.56 req/s success: 4960, failed: 0   // 48 loops in parallel for 3 min
  25.21 req/s success: 4538, failed: 0   // 64 loops in parallel for 3 min
  Running node lib/main load-test on ⬢ tc-queue-load-test... up, run.9665
  16.85 req/s success: 3033, failed: 0
  24.09 req/s success: 4336, failed: 0
  25.94 req/s success: 4670, failed: 0

createTask/claimWork/reportCompleted
  21.57 req/s success: 3883, failed: 0
  27.95 req/s success: 5031, failed: 0
  29.51 req/s success: 5311, failed: 0
  Running node lib/main.js load-test on ⬢ tc-queue-load-test... up, run.1628
  18.91 req/s success: 3403, failed: 0
  29.45 req/s success: 5301, failed: 0
  27.55 req/s success: 4959, failed: 0
  Running node lib/main.js load-test on ⬢ tc-queue-load-test... up, run.4989
  18.24 req/s success: 3284, failed: 0
  21.52 req/s success: 3874, failed: 0
  18.93 req/s success: 3408, failed: 0
```
So we have a through-put of about 20-25 tasks per second. I noticed that after a running the load-test a few times against the same process memory will spike and performance will degrade dramatically.
Maybe we have some leaks, but they seems to be related to load going up/down and then doing the same again... Sustained high load does not seems to cause the process to load memory... So perhaps it's some combination of bad luck or something..

Note the memory issue happens in both claimWork and the claimTask tests... It's trivially resolved by a restart. I suspect it's because we overload the processes and yeah, then something bad happens, I don't know... But it just needs time for the GC to catch up, or maybe we have leaks related to sockets or something.

Regardless I propose we land this.

imbstack:
## r+

Let's land this!

PR #89, opened 2016-04-21
Use cloud-mirror instead of s3-copy-proxy

jhford:


jonasfj:
Looks okay to me... Monitor it when you deploy... Please keep an eye on it and be ready to rollback... This is super critical.

jhford:
yep!  that's the plan

PR #90, opened 2016-05-16
Use new monitoring library instead of tc-lib-stats

imbstack:
This should Just Work™. I'm sure it won't for some reason and I'll roll back and push again, but it would be neat if it worked on the first try, right?


README.md
```diff
@@ -21,8 +21,8 @@ Development
 -----------
 
 To run tests you'll need a configuration file with access credentials for S3
-and Azure Blob and Table Storage, as well as pulse credentials and an influx db
-account. To do this, create a local configuration file
+and Azure Blob and Table Storage, as well as pulse credentials.
+To do this, create a local configuration file
 `taskcluster-queue.conf.json` in the root directory of the taskcluster-queue
```
jonasfj: he he this is so outdated it's bad :)

src/main.js
```diff
@@ -26,30 +25,12 @@ let load = base.loader({
...
-  },
   monitor: {
-    requires: ['cfg', 'influx', 'process'],
-    setup: ({cfg, influx, process}) => base.stats.startProcessUsageReporting({
-      drain:      influx,
-      component:  cfg.app.statsComponent,
-      process:    process
+    requires: ['profile', 'cfg'],
+    setup: ({profile, cfg}) => base.monitor({
+      project: 'taskcluster-queue',
```
jonasfj: qq, are we using this scheme elsewhere too? Just thinking `taskcluster-` might be a bit long.

Otherwise, nvm...

src/main.js
```diff
@@ -130,9 +109,7 @@ let load = base.loader({
           publicBucket:   ctx.publicArtifactBucket,
           privateBucket:  ctx.privateArtifactBucket
         },
-        drain:            ctx.influx,
-        component:        ctx.cfg.app.statsComponent,
-        process:          ctx.process
+        monitor: ctx.monitor.prefix('artifact'),
```
jonasfj: Maybe a pattern like `table.<tablename>` (ie. `table.artifacts`) is better?

Just because I can totally imagine us adding custom statistics about artifacts... Like size, etc... that isn't related to the azure table at all..

jonasfj: same goes for tasks...

src/main.js
```diff
@@ -431,46 +380,34 @@ let load = base.loader({
...
       var now = taskcluster.fromNow(cfg.app.taskExpirationDelay);
       assert(!_.isNaN(now), "Can't have NaN as now");
 
       // Expire task-requirement using delay
       debug("Expiring task-requirement at: %s, from before %s", new Date(), now);
       let count = await TaskRequirement.expire(now);
       debug("Expired %s task-requirement", count);
-
-      // Stop recording statistics and send any stats that we have
-      base.stats.stopProcessUsageReporting();
```
jonasfj: So in the old days did this some sort of CPU/memory monitoring... do we do any of that now?
If so, then don't we need to give it like a process identifier?
If not... then we probably shouldn't require `monitor` here, and add it once we have some sort of CPU/memory monitoring again...

jonasfj: @imbstack See:
 https://github.com/taskcluster/taskcluster-lib-stats/blob/090cb109b958d4579a7075eed62ac77817a94649/src/stats.js#L386-L431

Maybe `tc-lib-monitor` should have `monitor.resources(prefix)` where `prefix` will be the process id... like `web`, `background-worker`, `expire-task-requirement`, etc...

jonasfj: not saying this PR has to block on that..!

imbstack: Right now we do https://github.com/taskcluster/taskcluster-lib-monitor/blob/master/src/monitor.js#L36

I'm thinking that `monitor.resources()` would basically be regular prefixing, but it mutates the prefix of the currently running monitor, right?

src/main.js
```diff
@@ -431,46 +380,34 @@ let load = base.loader({
             new Date(), now);
       let count = await TaskGroupMember.expire(now);
       debug("Expired %s task-group members", count);
-
-      // Stop recording statistics and send any stats that we have
-      base.stats.stopProcessUsageReporting();
-      return influx.close();
```
jonasfj: Let's do "await monitor.flush()" here...

jonasfj: Additionally, first doing `monitor.count("expire-task-group-members.done", 1);` would make me very happy

PR #92, opened 2016-05-23
emit('error' when loops crash

jonasfj:


jonasfj:
@imbstack don't you think this will do?

In production nobody is listening for `on('error,... )` so it should be cause monitor to report to sentry and crash...

If I ever wanted to I could listen for the error in tests... though that won't be relevant..

imbstack:
r+ once the tests pass

src/dependencyresolver.js
```diff
@@ -88,9 +93,10 @@ class DependencyResolver {
         try {
           await this.dependencyTracker.resolveTask(m.taskId, m.resolution);
           await m.remove();
+          this.monitor.count('handled-messages-success', 1);
         } catch (err) {
-          debug("[alert-operator] Failed to handle message: %j" +
-                ", with err: %s, as JSON: %j", message, err, err, err.stack);
+          this.monitor.count('handled-messages-error', 1);
+          this.reportError(err, 'warning');
```
gregarndt: should this be this.monitor.reportError?

gregarndt: Does reporting an error cause things to crash then?

jonasfj: monitor.reportError is much better :)

jonasfj: > Does reporting an error cause things to crash then or is the goal to just catch this error, report it, and not crash?

Nope, but `.emit('error', err)` does... This is just a warning... the loop continues...

PR #93, opened 2016-05-23
Similar improements for other resolvers

jonasfj:


jonasfj:
Same as before, less critical... but let's get started... We really should add monitor here too, but let's not do that now :)

imbstack:
r+!

PR #94, opened 2016-05-23
List dependents

jonasfj:
This should enable us to list the dependents of a task...

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -102,22 +102,23 @@ var api = new base.API({
     InputError:       400,  // Any hand coded validation errors
   },
   context: [
-    'Task',
-    'Artifact',
-    'TaskGroup',
-    'taskGroupExpiresExtension',
-    'TaskGroupMember',
-    'publicBucket',
-    'privateBucket',
-    'blobStore',
-    'publisher',
-    'validator',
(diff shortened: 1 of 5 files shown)


src/api.js
```diff
@@ -102,22 +102,23 @@ var api = new base.API({
...
+    'TaskGroupMember',  // data.TaskGroupMember instance
+    'TaskDependency',   // data.TaskDependency instance
+    'publicBucket',     // bucket instance for public artifacts
+    'privateBucket',    // bucket instance for private artifacts
+    'blobStore',        // BlobStore for azure artifacts
+    'publisher',        // publisher from base.Exchanges
+    'validator',        // base.validator
+    'claimTimeout',     // Number of seconds before a claim expires
+    'queueService',     // Azure QueueService object from queueservice.js
+    'regionResolver',   // Instance of EC2RegionResolver,
```
imbstack: 🏆 for all of these comments!

imbstack:
r+ This looks solid to me. I wonder how it will perform for large numbers of descendants, but I guess there's only one way to find out!

jonasfj:
It does paging if there is more than 1k dependents... and it's not recursive...

PR #95, opened 2016-05-26
Must create tasks after their dependencies

djmitche:
I'm basing this determination on checks made in the `dependencyTracker`.


jonasfj:
# `r+`

jonasfj:
This is 100% correct... I thought it was a given...

PR #96, opened 2016-05-26
bug 1275326 fix createTask idempotency

jonasfj:
We don't need at-least one run anymore as createTask with task.dependencies as a self-loop is the same as defineTask, so no distinction


imbstack:
## r+

PR #97, opened 2016-05-27
Count poll-loop cycles

jonasfj:
Now we can make an alert in signalfx.. if we check that the loop counter is always above > 0.

This is similar to what jhford is doing with deadmansswitch...

### DIFF
--- a/src/dependencyresolver.js
+++ b/src/dependencyresolver.js
@@ -85,6 +85,8 @@ class DependencyResolver extends events.EventEmitter {
   async _pollResolvedTasks() {
     while(!this._stopping) {
       let messages = await this.queueService.pollResolvedQueue();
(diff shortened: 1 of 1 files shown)


gregarndt:
👍 💯

src/dependencyresolver.js
```diff
@@ -85,6 +85,8 @@ class DependencyResolver extends events.EventEmitter {
   async _pollResolvedTasks() {
     while(!this._stopping) {
       let messages = await this.queueService.pollResolvedQueue();
+      this.monitor.count('resolved-queue-poll-requests', 1);
```
gregarndt: just a fyi, but I think saying you're counting by one is superfluous here, I think it automatically uses one

jonasfj: true...

PR #98, opened 2016-06-28
More monitor

jonasfj:
I'll enable the lint test and fix issues with it later...

### DIFF
--- a/src/artifacts.js
+++ b/src/artifacts.js
@@ -409,8 +409,10 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
   // Handle azure artifacts
   if(artifact.storageType === 'azure') {
     if (artifact.details.container !== this.blobStore.container) {
-      debug("[alert-operator], Unknown container: %s, for artifact: %j",
-            artifact.details.container, artifact.json());
+      let err = new Error('Unknown container: ' +
+                          artifact.details.container + ' for artifact');
+      err.artifact = artifact.json();
+      await this.monitor.reportError(err);
     }
     // Generate URL expiration time
     var expiry = new Date();
@@ -436,8 +438,9 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
   }
 
   // We should never arrive here
-  debug("[alert-operator] Unknown artifact storageType from table storage: %s",
-        artifact.storageType);
+  let err = new Error('Unknown artifact storageType: ' + artifacts.storageType);
+  err.artifact = artifact.json();
+  this.monitor.reportError(err);
 };
 
 /** Get artifact from run */
--- a/src/bucket.js
+++ b/src/bucket.js
@@ -15,6 +15,7 @@ let Promise     = require('promise');
  *     secretAccessKey:  // ...
  *   },
  *   bucketCDN:          // https://cdn-for-bucket.com
+ *   monitor:            // base.monitor instance
  * }
  */
 var Bucket = function(options) {
@@ -23,11 +24,14 @@ var Bucket = function(options) {
   assert(options.credentials, "credentials must be specified");
   assert(!options.bucketCDN || typeof(options.bucketCDN) === 'string',
          "Expected bucketCDN to be a hostname or empty string for none");
+  assert(options.monitor,     "options.monitor is required");
   if (options.bucketCDN) {
     assert(/^https?:\/\//.test(options.bucketCDN), "bucketCDN must be http(s)");
     assert(/[^\/]$/.test(options.bucketCDN),
            "bucketCDN shouldn't end with slash");
   }
+  // Store the monitor
+  this.monitor = monitor;
   // Ensure access to the bucket property
   this.bucket = options.bucket;
   // Create S3 client
@@ -149,11 +153,10 @@ Bucket.prototype.setupCORS = async function() {
       debug("CORS already set for bucket: %s", this.bucket);
       return;
     }
-  }
-  catch (err) {
+  } catch (err) {
     // Failed to fetch CORS, ignoring issue for now
-    debug("[alert-operator] Failed to fetch CORS, err: %s, JSON: %j",
-          err, err, err.stack);
+    err.note = 'Failed to fetch CORS in bucket.js';
+    this.monitor.reportError(err, 'warning');
   }
 
   // Set CURS
--- a/src/bucket.js
+++ b/src/bucket.js
@@ -31,7 +31,7 @@ var Bucket = function(options) {
            "bucketCDN shouldn't end with slash");
   }
   // Store the monitor
-  this.monitor = monitor;
+  this.monitor = options.monitor;
   // Ensure access to the bucket property
   this.bucket = options.bucket;
   // Create S3 client
--- a/src/claimresolver.js
+++ b/src/claimresolver.js
@@ -157,7 +157,7 @@ class ClaimResolver {
     // report an error
     if (task.takenUntil.getTime() !== takenUntil.getTime()) {
       let err = new Error('Task takenUntil does not match takenUntil from ' +
-                          'message, taskId: ' + taskId ' this only happens ' +
+                          'message, taskId: ' + taskId + ' this only happens ' +
                           'if conditional load does not work');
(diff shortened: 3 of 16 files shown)


src/deadlineresolver.js
```diff
@@ -80,8 +81,13 @@ class DeadlineResolver extends events.EventEmitter {
...
     // Create promise that we're done looping
-    this.done = Promise.all(loops).catch((err) => {
-      this.emit('error', err); // This should crash the process
+    this.done = Promise.all(loops).catch(async (err) => {
+      console.log("Crashing the process: %s, as json: %j", err, err);
+      // TODO: use this.monitor.reportError(err); when PR lands:
+      // https://github.com/taskcluster/taskcluster-lib-monitor/pull/27
+      await this.monitor.reportError(err, 'error', {}, true);
+      // Crash the process
+      process.exit(1);
```
djmitche: It seems like this behavior should be factored out to somewhere common, perhaps with tests?

src/artifacts.js
```diff
@@ -436,8 +438,9 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
   }
 
   // We should never arrive here
-  debug("[alert-operator] Unknown artifact storageType from table storage: %s",
-        artifact.storageType);
+  let err = new Error('Unknown artifact storageType: ' + artifacts.storageType);
+  err.artifact = artifact.json();
+  this.monitor.reportError(err);
```
gregarndt: should wait for this promise to resolve?

src/bucket.js
```diff
@@ -23,11 +24,14 @@ var Bucket = function(options) {
   assert(options.credentials, "credentials must be specified");
   assert(!options.bucketCDN || typeof(options.bucketCDN) === 'string',
          "Expected bucketCDN to be a hostname or empty string for none");
+  assert(options.monitor,     "options.monitor is required");
```
gregarndt: nit, this error doesn't is a little different than the others... perhaps just "monitor must be given" ?  I'm using the bucket and credentials assertions above as an example.

src/bucket.js
```diff
@@ -149,11 +153,10 @@ Bucket.prototype.setupCORS = async function() {
...
       return;
     }
-  }
-  catch (err) {
+  } catch (err) {
     // Failed to fetch CORS, ignoring issue for now
-    debug("[alert-operator] Failed to fetch CORS, err: %s, JSON: %j",
-          err, err, err.stack);
+    err.note = 'Failed to fetch CORS in bucket.js';
+    this.monitor.reportError(err, 'warning');
```
gregarndt: need to wait for this to complete?

src/claimresolver.js
```diff
@@ -28,7 +28,7 @@ const RESOLVED_STATES = [
  * ignored as it implies a failed request (which wasn't retried) or a run
  * that was reclaimed.
  */
-class ClaimResolver extends events.EventEmitter {
+class ClaimResolver {
```
gregarndt: if this no longer extends EventEmitter, do you need to import events anymore?

src/claimresolver.js
```diff
@@ -57,12 +56,14 @@ class ClaimResolver extends events.EventEmitter {
            "Expected pollingDelay to be a number");
     assert(typeof(options.parallelism) === 'number',
            "Expected parallelism to be a number");
+    assert(options.monitor !== null, 'options.monitor required!');
```
gregarndt: same comment I gave with the other assertion

src/claimresolver.js
```diff
@@ -102,13 +108,14 @@ class ClaimResolver extends events.EventEmitter {
...
         // Don't let a single task error break the loop, it'll be retried later
         // as we don't remove message unless they are handled
-        return this.handleMessage(message).catch((err) => {
-          debug("[alert-operator] Failed to handle message: %j" +
-                ", with err: %s, as JSON: %j", message, err, err, err.stack);
-        });
+        try {
+          await this.handleMessage(message);
+        } catch (err) {
+          this.monitor.reportError(err, 'warning');
```
gregarndt: Before you were including the message in the alert-operator statement, do you not want to include that in the error you're reporting?

src/claimresolver.js
```diff
@@ -191,8 +200,11 @@ class ClaimResolver extends events.EventEmitter {
 
       // If the run isn't the last run, then something is very wrong
       if (task.runs.length - 1 !== runId) {
-        debug("[alert-operator] running runId: %s, resolved exception, " +
-              "but it wasn't the last run! taskId: ", runId, taskId);
+        let err = new Error('Running runId: ' + runId + ' resolved exception,' +
+                            'but it was not the last run! taskId: ' + taskId);
+        err.taskId = taskId;
+        err.runId = runId;
+        this.monitor.reportError(err);
```
gregarndt: await?

src/data.js
```diff
@@ -316,9 +317,12 @@ Artifact.prototype.remove = function(ignoreError) {
...
     } else {
-      debug("[alert-operator] Expiring artifact with bucket: %s, which isn't " +
-            "configured for use. Please investigate taskId: %s, runId: %s",
-            this.details.bucket, this.taskId, this.runId);
+      let err = new Error("Expiring artifact with bucket which isn't " +
+                          "configured for use. Please investigate!");
+      err.bucket  = this.details.bucket;
+      err.taskId  = this.taskId;
+      err.runId   = this.runId;
+      this.monitor.reportError(err);
```
gregarndt: await?

src/data.js
```diff
@@ -316,9 +317,12 @@ Artifact.prototype.remove = function(ignoreError) {
     } else if (this.details.bucket === this.privateBucket.bucket) {
       deleted = this.privateBucket.deleteObject(this.details.prefix);
     } else {
-      debug("[alert-operator] Expiring artifact with bucket: %s, which isn't " +
-            "configured for use. Please investigate taskId: %s, runId: %s",
-            this.details.bucket, this.taskId, this.runId);
+      let err = new Error("Expiring artifact with bucket which isn't " +
```
gregarndt: nit: you intermingle single and double quotes when you deal with strings, I suppose the linter stuff will pick up on this if there is the right rule for it.

src/data.js
```diff
@@ -329,9 +333,12 @@ Artifact.prototype.remove = function(ignoreError) {
...
     if (this.details.container !== this.blobStore.container) {
-      debug("[alert-operator] Expiring artifact with container: %s, which " +
-            "configured for use. Please investigate taskId: %s, runId: %s",
-            this.details.container, this.taskId, this.runId);
+      let err = new Error("Expiring artifact with container which isn't " +
+                          "configured for use. Please investigate!");
+      err.container = this.details.container;
+      err.taskId    = this.taskId;
+      err.runId     = this.runId;
+      this.monitor.reportError(err);
```
gregarndt: ok, so I think that this should be used with "await" but I'll refrain from commenting on it anymore just to reduce the noise.

src/deadlineresolver.js
gregarndt: +1, especially since this piece I think has bitten us twice now.

gregarndt:
Took a look and I think the larged chunk of comments is regarding awaiting for the resolution of the reportError() promise.

src/claimresolver.js
jonasfj: good point

src/data.js
jonasfj: linter PR will fix this...

jonasfj:
I have mixed feelings with regards to waiting for the `monitor.reportError` to resolve...

I see error reporting as a best effort service... So I'm not sure there is any need to wait for it in most cases... I do it if the error is really critical...

But it's mostly for cases where I want to shutdown after reporting the error...

jonasfj:
Does that make sense?

gregarndt:
makes sense, thanks.

src/deadlineresolver.js
jonasfj: yeah, that would be nice... It won't be in this PR though...
Ideally, some generic way of constructing a process that polls from azure queue and operates on the messages...

PR #99, opened 2016-06-28
Fixed a lot of linter issues

jonasfj:
Fixed all the linter issues...


gregarndt:
I just gave this a quick glance since it's mostly fixing small things to make the linter happy.  I think that as long as no logic changes in this PR, the tests pass, then it's good to go.

jonasfj:
Awesome I'll take that as an r+

I won't land until I have an r+ for: https://github.com/taskcluster/taskcluster-queue/pull/98
which this is based on...

PR #103, opened 2016-07-20
Add pulse message for Task Group Resolved

imbstack:
I still want to add more tests, come up with some nicer variable/object names, and figure out if edge cases are covered here, but I think this seems like it works. Does this seem like it accomplishes what we talked about last week?


config.yml
```diff
@@ -18,6 +18,7 @@ defaults:
     artifactTableName:            Artifacts
     taskGroupTableName:           TaskGroups
     taskGroupMemberTableName:     TaskGroupMembers
+    taskGroupSizeTableName:       TaskGroupSizes
```
jonasfj: hmm... Maybe we can it something `taskGroupActiveSetTableName`  ?? Just brain storming...

src/api.js
```diff
@@ -486,6 +487,18 @@ let ensureTaskGroup = async (ctx, taskId, taskDef, res) => {
...
+  // Now we walso add the task to the group size counters as well
+  await ctx.TaskGroupSize.create({
+    taskGroupId,
+    taskId,
+    expires: new Date(taskDef.expires),
+  }).catch(err => {
+    // If the entity already exists, then we're happy no need to crash
+    if (!err || err.code !== 'EntityAlreadyExists') {
+      throw err;
+    }
```
jonasfj: We need some check of taskDef.expires here... I suspect it has to exactly the same... as that is the check we do for  `task.Create`... see `_.isEqual(taskDef, def)`

src/api.js
```diff
@@ -486,6 +487,18 @@ let ensureTaskGroup = async (ctx, taskId, taskDef, res) => {
     }
   });
 
+  // Now we walso add the task to the group size counters as well
+  await ctx.TaskGroupSize.create({
```
jonasfj: I'm not sure this is the ideal place to do this...
Here is the current process:
 1) TaskGroupMember.create(...)
 2) putDeadlineMessage
 3) TaskGroupSize.create(...)
 4) Task.create(...)
 5) Track dependencies (if needed)
 6) Pulse taskDefined
 7) putPendingMessage (if needed) 
 8) Pulse taskPending (if needed)

This ordering isn't arbitrary, we handle failure modes here... When we do listTaskGroup, we query the `TaskGroupMember` table, but for every entry we lookup the `Task` table to ensure the task was in fact created. When handling deadline messages we ignore cases where the `Task` table entry is missing, because the node could have crashed in the process of making it...

I would suggest that we insert in the `TaskGroupSize` table after `putDeadlineMessage`... Probably before `Task.Create(...)`... definitely before we track dependencies or `putPendingMessage`... So step 3.
Then we patch deadline resolution logic to delete the `TaskGroupSize` even if there was no entry in the `Task` table.

Reasoning:
If `queue.createTask` fails and isn't retried... Then if it crashes before:
 * step 2: Nothing happens (TaskGroupMember refres to Task that doesn't exist, the list logic will filter it out)
 * step 3: deadline message refers Task and TaskGroupSize entries that doesn't exist, nothing happens.
 * step 4: taskGroup-resolved message is delayed until the task.deadline when the deadline message, sees that the Task entry isn't defined, but the TaskGroupSize is defined... So it deletes TaskGroupSize and sends the message.
 * step 5: task completed message and taskGroupId resolved messages are delayed until task.deadline. Because even though the task was created, dependencies aren't tracking it and pending message wasn't sent... But deadline message will resolve it.
....

Notice that this way we always end up in a sane state...  Sometimes we have tasks (and task-groups) that weren't created successfully, and retries weren't followed up. The result would be that such tasks and task-groups are eventually resolved by deadline expiration. They simplify fail to become pending (and doesn't get picked up by a worker, even if they were initially pending).

src/dependencytracker.js
```diff
@@ -252,6 +255,30 @@ class DependencyTracker {
...
+    let remaining = -1;
+
+    await this.TaskGroupSize.query({taskGroupId: task.taskGroupId}, {
+      handler: async (sizeCounter) => {
+        remaining += 1;
+        if (sizeCounter.taskId === taskId) {
+          await sizeCounter.remove();
+        }
+      },
+    });
```
jonasfj: Not the right logic...
Use:
```js
// remove entry, ignore if it doesn't exist (that's that true does, I think... check code comments in azure-entities)
await this.TaskGroupSize.remove({taskId: task.taskId, taskGroupId: task.taskGroupId}, true});

// check for emptyness of the partition
let result = await this.TaskGroupSize.query({taskGroupId: task.taskGroupId}, {limit: 1});

// We assume this won't happen
assert(result.entries.length > 0 || !result.continuation, "hmm... this is bad, see comment in isBlocked()");

if (result.entries.length == 0) {
  // done...
}
```

src/dependencytracker.js
```diff
@@ -252,6 +255,30 @@ class DependencyTracker {
         }
       },
     });
+
+    let task = await this.Task.load({taskId: taskId});
```
jonasfj: I suggest checking the call site for `resolveTask` I'm sure it already has the task entity... so we can probably just take taskGroupId as an other option....

If there are places where it isn't present we can take it as an optional argument... we only do the Task lookup here because we need the taskGroupId, right?

imbstack: I think the most common call site for this is from dependencyResolver which doesn't have taskGroupId or load a Task before calling dependencyTracker.

That said, the places that call `putResolvedMessage` both have task loaded and ready, so we can just add it to the message in the azure queue, right? What's the migration path for that look like? Is there a scheme associated with it and such or do we just need to be able to handle both message formats for a bit and all will work out ok?

jonasfj: > do we just need to be able to handle both message formats for a bit and all will work out ok?

Yeah that would do...there is no schema...

I guess if we just make `taskGroupId` an optional parameter, and load if it not present it'll just work when we pass `m.taskGroupId` as parameter... It'll obviously be undefined for messages that doesn't have a taskGroupId.

You you can just land a quick hack now that add taskGroupId to all messages... that should be a quick thing... then you can also run npm shrinkwrap again and get the new API and azure-entities in there :)

src/exchanges.js
```diff
@@ -297,3 +297,23 @@ exchanges.declare({
...
+    'when a group is "complete". There is no guarantee that this group is now',
+    'intert. It can have more tasks added to it by the scheduler, but at least',
+    'for the time being, it is done. If another task is added to this group and',
+    'it completes, this message will be sent again.',
+  ].join('\n'),
+  routingKey:         buildCommonRoutingKey(),
+  schema:             'task-group-resolved.json#',
+  messageBuilder:     commonMessageBuilder,
+  routingKeyBuilder:  commonRoutingKeyBuilder,
+  CCBuilder:          commonCCBuilder,
```
jonasfj: Don't use this, this isn't cc'ed to anything

src/exchanges.js
```diff
@@ -297,3 +297,23 @@ exchanges.declare({
...
+    'a message will be sent on this exchange. It can be listened to to know',
+    'when a group is "complete". There is no guarantee that this group is now',
+    'intert. It can have more tasks added to it by the scheduler, but at least',
+    'for the time being, it is done. If another task is added to this group and',
+    'it completes, this message will be sent again.',
+  ].join('\n'),
+  routingKey:         buildCommonRoutingKey(),
+  schema:             'task-group-resolved.json#',
+  messageBuilder:     commonMessageBuilder,
+  routingKeyBuilder:  commonRoutingKeyBuilder,
```
jonasfj: Not sure how much of this you can use...

jonasfj: isn't it unfair to include taskId from the last task... when this is really about the task-group...

src/api.js
imbstack: Would you rather have us load a task here or take it as an argument to the method? I expect most places that call this method will already have the task loaded, but perhaps not?

jonasfj: why would you load the task?

You just need to check expiration... and you already have that... if the expiration match, then continue... otherwise done... 

If the task is already created, it'll be checked later...

schemas/poll-task-urls-response.yml
```diff
@@ -64,6 +64,7 @@ properties:
       expires and not longer works for authentication.
     type:         string
     format:       date-time
+additionalProperties: false
```
jonasfj: wow, that's been missing many places... ++++

src/api.js
```diff
@@ -478,14 +479,44 @@ let ensureTaskGroup = async (ctx, taskId, taskDef, res) => {
...
+    // If the entity already exists, then we're happy no need to crash
+    if (!err || err.code !== 'EntityAlreadyExists') {
+      throw err;
+    }
+
+    let active = await ctx.TaskGroupActiveSet.load({taskId, taskGroupId});
+
+    if (!_.isEqual(new Date(active.expires), expires)) {
+      return res.reportError('RequestConflict', [
+        'TaskGroupActiveSet for taskId:{{taskId}} taskGroupId:{{taskGroupId}}',
```
jonasfj: Don't mention `TaskGroupActiveSet`, just say `taskId` already in use... `TaskGroupActiveSet` is an internal thing...

jonasfj: Just say `taskId` already exists, and the task has expiration that is different...

src/dependencytracker.js
```diff
@@ -252,6 +255,20 @@ class DependencyTracker {
...
       },
     });
+
+    await this.TaskGroupActiveSet.remove({taskId, taskGroupId}, true);
+
+    // check for emptiness of the partition
+    let result = await this.TaskGroupActiveSet.query({taskGroupId}, {limit: 1});
+
+    // We assume this generally won't happen
+    assert(result.entries.length > 0 || !result.continuation, 'hmm... this is bad, see comment in isBlocked()');
```
jonasfj: Actually, couldn't we do `new Error(...)` and then add the `taskId` and `taskGroupId` as properties on the error.. that way we'll have them in sentry... it's only a few lines...

src/deadlineresolver.js
```diff
@@ -208,7 +208,7 @@ class DeadlineResolver {
       debug('Resolved taskId: %s, by deadline', taskId);
 
       // Update dependency tracker
-      await this.dependencyTracker.resolveTask(taskId, 'exception');
+      await this.dependencyTracker.resolveTask(taskId, task.taskGroupId, 'exception');
```
jonasfj: Further up in this file... if `if (!task) {` and we return immediately, we need to remove any partially constructed task...

Basically, we'll have to call `this.TaskGroupActiveSet.remove({taskId, taskGroupId}, true);` and check for emptiness and send pulse message if empty...

The `if (!task` case happens if the task wasn't created by the deadline message was sent... Due to a crash... In that case the `TaskGroupActiveSet` entry might have been created... We should make sure it's cleaned up... And that the pulse message is sent...

So failing to retry a createTask (which is what can cause this), will result in a task-group something that's resolved by deadline... I can live with that... Obviously, that's an extremely delayed message, but a lot better than nothing.

src/dependencytracker.js
jonasfj: See PR #107

src/deadlineresolver.js
imbstack: If we don't have a task, we won't be able to check for an empty TaskGroupActiveSet because we don't know the task-group of the non-existent task, right? I can totally see why it's nice to remove the TaskGroupActiveSet entry, but we can't really check for the emptiness, right?

jonasfj: @imbstack I think this is a case where we have to add it to the message again...

The deadline message serves 3 purposes:
 A) Resolve old tasks
 B) Resolve tasks that failed to create properly
 C) Clean up TaskGroupActiveSet and publish message about task-group finished...

jonasfj: @imbstack, while doing this... it might be a good idea of you write down some comments... for example when sending the deadline message documenting what it responsible for doing...

I'm not sure we'll be able to remember all this next time we have to look at it :)

jonasfj:
# `r+`

src/exchanges.js
```diff
@@ -297,3 +317,23 @@ exchanges.declare({
...
+/** Task group resolved exchange */
+exchanges.declare({
+  exchange:           'task-group-resolved',
+  name:               'taskGroupResolved',
+  title:              'Task Group Resolved',
+  description: [
+    'Whenever a task group has run to completion (success is not considered),',
+    'a message will be sent on this exchange. It can be listened to to know',
+    'when a group is "complete". There is no guarantee that this group is now',
+    'intert. It can have more tasks added to it by the scheduler, but at least',
```
petemoore: @imbstack What is intert? I guess this is a typo, but I can't work out what word should be there. :-)

Inert?

Also, I thought we were ditching the scheduler - are tasks not added now by the queue?

PR #105, opened 2016-07-23
Count when work queues reach zero

jonasfj:
Some stats to see if claimResolvers work queue reaches zero... If it doesn't that explains why @lundjordan saw a task in state `running` 14 hours after `takenUntil` expired.


imbstack:
### r+

PR #106, opened 2016-07-26
bug 1285506 -- let queue rewrite urls for eu-central-1

jhford:
Since we're no longer considering the useCloudMirror flag in the task, let's not bother loading it any longer.  This removes some more s3-copy-proxy leftovers and cleans up the configs.  It also adds eu-central-1 to the list of regions that we rewrite urls for, since we now have this regions in cloud-mirror

@jonasfj, I don't have a full test environment for this, can you double check that it looks OK?


jonasfj:
Looks fine to me... landing,

PR #109, opened 2016-07-29
Add schedulerId to some azure message queues

imbstack:
As we discussed yesterday. I can also do a shrinkwrap update on this. Give me a sec to push that as well and we can deploy at once.

### DIFF
--- a/test/queueservice_test.js
+++ b/test/queueservice_test.js
@@ -50,7 +50,7 @@ suite('queue/QueueService', function() {
   test('putDeadlineMessage, pollDeadlineQueue', async () => {
     var taskId      = slugid.v4();
     var taskGroupId = slugid.v4();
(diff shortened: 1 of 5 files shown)


test/queueservice_test.js
```diff
@@ -50,10 +50,11 @@ suite('queue/QueueService', function() {
   test('putDeadlineMessage, pollDeadlineQueue', async () => {
     var taskId      = slugid.v4();
     var taskGroupId = slugid.v4();
+    var schedulerId = 'fake-garbage-scheduler';
```
jonasfj: Use a slugid instead... they are designed to fit these generic identifiers.. .and we reduce conflicts with tests running in parallel (we do have other conflicts)

jonasfj: or maybe it's not important here...

jonasfj:
# `r+`

PR #110, opened 2016-07-29
Start writing schedulerId in task-group-resolved

imbstack:
As we discussed in https://public.etherpad-mozilla.org/p/tc-task-group-queue-stuff


jonasfj:
# `r+`


You might want to land changes to the messages first...

Also please consider adding the negative test case for checking if a pulse message will be published too soon. It's nice to have for future proofing this...

PR #111, opened 2016-07-29
Set output: blob

jonasfj:


imbstack:
## r+

djmitche:
Yikes, this is ancient.  Do we still support output; 'blob', @jonasfj?  If so, perhaps a new implementation of this is in order!

djmitche:
Looks like it's implemented in tc-lib-api, but not documented.  I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1453719 to finish this up.

PR #113, opened 2016-08-12
Update monitor to not loop on failures

imbstack:
As it says on the tin.

This also updates tc-base to be a version that includes lib-iterate for when we work on that.


jonasfj:
# `r+`

PR #114, opened 2016-08-23
Description of task expiraiton has a missing word.

SingingTree:
The description of task expiration looks like it's missing 'task' in the
sentence: "Notice that all artifacts for the must have an expiration that is no
later than this."


jonasfj:
Awesome... Thanks!

# `r+`

PR #116, opened 2016-09-01
Add inverse test for task group resolved

imbstack:
Also make schedulerId mandatory for task-group-resolved messages while I'm in here.

### DIFF
--- a/test/taskgroup_test.js
+++ b/test/taskgroup_test.js
@@ -84,8 +84,11 @@ suite('TaskGroup features', () => {
       workerGroup:    'my-worker-group',
       workerId:       'my-worker',
     });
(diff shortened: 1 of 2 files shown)


test/taskgroup_test.js
```diff
@@ -79,8 +85,9 @@ suite('TaskGroup features', () => {
       workerId:       'my-worker',
     });
     await helper.queue.reportCompleted(taskIdB, 0);
+    shouldveResolved = true;
```
jonasfj: must set true before you reportCompleted, this test is intermittet

jonasfj: but you can sleep 100ms before doing so...

jonasfj:
# `r+`

PR #117, opened 2016-09-12
non-printable ascii in artifact names -> 400

jonasfj:
See: https://sentry.io/taskcluster/taskcluster-queue/issues/158299920/


imbstack:
## r+

nice find

PR #120, opened 2016-09-28
add request-region endpoint to preview queue-region

jhford:
Basically, just add a diagnostic endpoint to see which region the queue thinks that a requester lives in, or not a region.  There are three linter errors here locally, but they are in a file that I didn't touch.  @jonasfj, if you think this is OK, can you deploy it?

### DIFF
--- a/src/artifacts.js
+++ b/src/artifacts.js
@@ -460,6 +460,20 @@ api.declare({
   return replyWithArtifact.call(this, taskId, runId, name, req, res);
 });
 
+/** Tell me which region the queue thinks that requester is in */
+api.declare({
+  method: 'get',
+  route: '/request-region',
+  name: 'get-request-region',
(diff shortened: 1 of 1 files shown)


jonasfj:
@jhford, what do you need this for?

If you want to know what region it resolves you to. You can just do a request without redirects and read the `location` header.

`curl -v https://queue...`

I'm not a fan of polluting the API with this sort of thing.
If you wish to land it temporarily and then back it out that's okay, but really... Just test it using curl without `-L` it won't follow redirects.

djmitche:
@jonasfj it's to help debug https://bugzilla.mozilla.org/show_bug.cgi?id=1306040.  I imagine it would be removed once that's debugged.

petemoore:
No longer needed.

PR #123, opened 2016-10-04
Fixed wording of deadline description

petemoore:
Deadline exceeded is an exception not a failure...


petemoore:
Given that we raised identical code changes in PR #122 and PR #123 I'm assuming you're happy with my changes! :-)

PR #124, opened 2016-10-14
Added hintId to all pending azure queue messages

jonasfj:
Once this is deployed we can modify the claimWork logic to ensure that we record the hintId for each claim...
Notably we can use this ensure that two different azure messages doesn't lead to the same task being claimed twice by the same worker (when there is a duplication of azure messages)...

Note: this could break workers if they won't ignore extra properties...
@petemoore ^ do you know if additional properties will break generic-worker, I doubt it.. but it would be hard to back out...


jonasfj:
@petemoore, gave an r+ for this on IRC...

jonasfj:
as per IRC discussion with pete this should be perfectly safe to deploy...

PR #126, opened 2016-10-14
Use hint

jonasfj:
This is solve the issue garndt found where claimWork return two claims for the same task...

Note: we can't land this without #124, which should have be deployed at least 24 hours before we deploy this...


jonasfj:
We can deploy this in a few days... #124  was just deployed.

PR #127, opened 2016-10-21
New ping API

owlishDeveloper:
@djmitche @imbstack r?

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "superagent-promise": "^0.2.0",
     "taskcluster-base": "^6.0.0",
     "taskcluster-client": "^0.23.8",
+    "taskcluster-lib-api": "^3.0.0",
     "tc-rules": "^5.0.0",
     "thirty-two": "^0.0.2",
     "url-join": "^0.0.1",
--- a/src/api.js
+++ b/src/api.js
@@ -3,6 +3,7 @@ let debug       = require('debug')('app:api');
 let slugid      = require('slugid');
 let assert      = require('assert');
 let _           = require('lodash');
+let API         = require('taskcluster-lib-api');
 let base        = require('taskcluster-base');
 let taskcluster = require('taskcluster-client');
 
@@ -74,7 +75,7 @@ var RUN_ID_PATTERN      = /^[1-9]*[0-9]+$/;
  *   workClaimer:    // WorkClaimer instance from workclaimer.js
  * }
  */
-var api = new base.API({
+var api = new API({
   title:        'Queue API Documentation',
   description: [
     'The queue, typically available at `queue.taskcluster.net`, is responsible',
@@ -133,7 +134,7 @@ api.declare({
   method:     'get',
   route:      '/task/:taskId',
   name:       'task',
-  stability:  base.API.stability.stable,
+  stability:  API.stability.stable,
   idempotent: true,
   output:     'task.json#',
   title:      'Get Task Definition',
@@ -169,7 +170,7 @@ api.declare({
   method:     'get',
   route:      '/task/:taskId/status',
   name:       'status',
-  stability:  base.API.stability.stable,
+  stability:  API.stability.stable,
   input:      undefined,  // No input is accepted
   output:     'task-status-response.json#',
   title:      'Get task status',
@@ -207,7 +208,7 @@ api.declare({
     limit: /^[0-9]+$/,
   },
   name:       'listTaskGroup',
-  stability:  base.API.stability.stable,
+  stability:  API.stability.stable,
   output:     'list-task-group-response.json#',
   title:      'List Task Group',
   description: [
@@ -290,7 +291,7 @@ api.declare({
     limit: /^[0-9]+$/,
   },
   name:       'listDependentTasks',
-  stability:  base.API.stability.stable,
+  stability:  API.stability.stable,
   output:     'list-dependent-tasks-response.json#',
   title:      'List Dependent Tasks',
   description: [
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -37,6 +37,7 @@
     "superagent-promise": "^0.2.0",
     "taskcluster-base": "^6.0.0",
     "taskcluster-client": "^0.23.8",
+    "taskcluster-lib-api": "^3.0.0",
```
djmitche: This won't be enough, as `src/api.js` uses `base.API`, thus getting whatever version of taskcluster-lib-api is required by taskcluster-base.

That will need to be modified to `require('taskcluster-lib-api')` instead.

owlishDeveloper: Done!

src/api.js
```diff
@@ -3,7 +3,7 @@ let debug       = require('debug')('app:api');
 let slugid      = require('slugid');
 let assert      = require('assert');
 let _           = require('lodash');
-let base        = require('taskcluster-base');
+let base        = require('taskcluster-lib-api');
```
owlishDeveloper: Done!

owlishDeveloper: I noticed in the comments references to base, for example: https://github.com/owlishDeveloper/taskcluster-queue/blob/a2ed15fd0b144ca332396a3b928359d7dbfcafda/src/api.js#L117 - does it mean, if I deleted base, I should require, in this case, pulse-publisher (I saw in tc-base `Exchanges:      'pulse-publisher'`)? Also, there were couple of base.Entity, I changed that as well, but I'm not sure if that's correct...

djmitche: Yes, upgrading from base is tricky, because you don't really know what version of the libraries you're upgrading from.  Typically I've just required the one library I need (in this case, tc-lib-api) and left everything else pointing at base.  That said, upgrading everything is worth a try here, and as long as it passes tests and gets @jonasfj's :+1: I'm more than happy to see it done.

package.json
djmitche: I'd rather leave base where it is

owlishDeveloper:
Interestingly enough, it passes the tests as it is. @djmitche , from your comments I see there are two options:
1. I leave the base for Entity in api.js
2. I remove base from queue completely, like I did in tc-index and aws-provisioner yesterday (a lot of changes in multiple src and test files, and in package.json). You said you would be more than happy to see this done, but then you suggested the first option, so I'm confused :) What would rather have me do? I guess this question goes also to @jonasfj

djmitche:
It's a matter of balancing risk against reward.  If we break index, take an hour to realize it, and roll back, it's not too bad.  But if we break queue, lots of bad things happen.  So the risk is higher.  I think you're right, though -- if @jonasfj is happy with the change and willing to do a careful deployment and watch for issues, I'm OK with forging on ahead.  I've wanted to get rid of base for a long time :)

owlishDeveloper:
Well, I'm up! So, the decision is yours, @jonasfj

jonasfj:
`npm-shrinkwrap.json` needs to be present... So please re-add it :)
My flow is usually: `rm -rf node_modules npm-shrinkwrap.json && npm install && npm shrinkwrap && git commit -am 'updated shrinkwrap'` 

If you have tests passing I'll be happy to merge this... If you are into removing tc-base everywhere that would be cool too.. IMO, just do in another PR to keep things light.

jonasfj:
Sorry for the slow turnaround...

owlishDeveloper:
@jonasfj OK, I'll make a separate PR for tc-base! Thank you!

package.json
```diff
@@ -13,6 +13,8 @@
     "url": "https://github.com/taskcluster/taskcluster-queue.git"
   },
   "dependencies": {
+    "ajv": "^4.2.0",
+    "ajv-keywords": "^1.1.1",
```
jonasfj: Why do we need them here?
Shouldn't they be dependencies of `taskcluster-lib-validate`

And wouldn't it be enough to list `ajv-keywords` in-order to satisfy the peerDependency?

Hmm... If this is a new change in ajv, perhaps we should file a bug with ajv, given that adding a peerDependency is a breaking change. So they should have upgraded to a new major version.
Before we file that we should make sure that ajv did in fact change from not having a peerDep to having a peerDep in the 4.x series.

jonasfj: hmm... we really should find out why or what is wrong here.. Adding dependencies like this isn't a solution as it'll spread to all other components and become unmaintainable... Maybe one of the upstream packages are missing a dependency...

owlishDeveloper: I'll check!

owlishDeveloper: @jonasfj So, `taskcluster-lib-api` wants `^3.8.5`. That happens to be the highest level in the tree (`taskcluster-lib-api` is queue's dependency). So, `ajv` is installed in queue's `node_modules`, and the version is 3.8.10. Now, the table has its own `ajv`, which is 4.8.2. It doesn't have its own `ajv-keywords`, however. As per default npm 3's behavior, `ajv-keywords` goes to the queue's modules. What seems to be happening, is that `ajv-keywords` uses the `ajv` in the same directory as it self, not the one in the `table`'s npm_modules. It probably happens because `ajv` is a peer dependency, not regular dependency. (I assume)

We could probably use a newer `ajv` in `tc-lib-api`, that could be a solution.

imbstack: Wow. I really don't like npm sometimes. Nice sleuthing, @owlishDeveloper.

owlishDeveloper:
The problem of peer dependencies is now resolved, everything is being installed properly. I was able to write shrinkwrap. @jonasfj r?

jonasfj:
cool, let's see what happens it should auto-deploy once tests have passed.

PR #128, opened 2016-10-25
NO MERGE: Proof-of-concept showing neutrino, TC preset, and Yarn

eliperelman:
DO NOT MERGE!

This PR is meant as a proof of concept. It demonstrates using neutrino and neutrino-preset-taskcluster out of the box. It also demos using Yarn to eliminate npm shrinkwrap issues.


src/index.js
```diff
@@ -516,16 +515,5 @@ let load = base.loader({
 
 }, ['profile', 'process']);
 
-// If this file is executed launch component from first argument
-if (!module.parent) {
-  load(process.argv[2], {
```
jonasfj: I'm pretty sure this part would still have been needed :)

test/queueservice_test.js
```diff
@@ -2,7 +2,7 @@ suite('queue/QueueService', function() {
   var Promise       = require('promise');
   var slugid        = require('slugid');
   var assert        = require('assert');
-  var QueueService  = require('../lib/queueservice');
+  var QueueService  = require('../src/queueservice');
```
jonasfj: Does neutrino do path rewriting?

src/index.js
eliperelman: This was moved to root/index to simplify the call and separate the bin from the module.

test/queueservice_test.js
eliperelman: It can, but it doesn't. Using mocha and babel with Node builds the source files.

PR #129, opened 2016-10-31
Add support for x-taskcluster-skip-cache header

jonasfj:


src/artifacts.js
```diff
@@ -446,6 +449,13 @@ api.declare({
     'stored externally. Either way, the response may not be JSON. So API',
     'client users might want to generate a signed URL for this end-point and',
     'use that URL with a normal HTTP client.',
+    '',
+    '**Caching**, artifacts may be cached in data centers closer to the',
+    'workers in-order to reduce bandwidth costs. This can lead to response',
+    'times at upwards 30 seconds. Caching can be skipped by setting the header',
```
jhford: I don't think we should be putting a specific time in this comment.  Let's leave it as "This can lead to longer response times when there are problems with resource caching".

src/artifacts.js
```diff
@@ -356,6 +356,9 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
       if (!region) {
         debug('artifact from CDN for ip: %s', req.headers['x-forwarded-for']);
         url = this.publicBucket.createGetUrl(prefix);
+      } else if (req.headers['x-taskcluster-skip-cache'] === 'true') {
```
jhford: should we `.toLowerCase()` the header value before comparison?  Maybe we should also support `1|0`?

gregarndt:
I'm +1 for this.  In the case of resources in AWS where this would be used (mainly Treeherder), it's either we incur a transfer by cloud-mirror and then serve from that bucket, or we direct Treeherder to the original bucket still incurring the transfer.  That seems acceptable in this case.

PR #132, opened 2016-11-16
Add tc-lib-docs

imbstack:
As it says on the tin.


src/main.js
```diff
@@ -61,6 +62,32 @@ let load = loader({
     }),
   },
 
+  reference: {
```
jonasfj: Why do we need this level of indirection? `reference()` is not asynchronous... and even if it was, we could just make the `setup` for `docs` component async and it would be fine...

jonasfj: https://github.com/taskcluster/taskcluster-queue/pull/133

PR #133, opened 2016-11-16
Removing indirection

jonasfj:
A PR for your PR.


imbstack:
I don't remember why we didn't do it this way in the first place...

PR #135, opened 2016-12-15
Increase route limit

jonasfj:


jonasfj:
Fixing [bug 1323792](https://bugzilla.mozilla.org/show_bug.cgi?id=1323792)

PR #137, opened 2016-12-31
add schema docs

djmitche:
This isn't optimal.  Before https://github.com/taskcluster/taskcluster-docs/pull/142/commits/ea7386919b925e1f39f2a3cdbdf8d76043d12bc4, there was an extra API reference page like this one showing just the task schema.  tc-lib-docs doesn't provide a way to include such a thing.  So we'll put it under docs.

I'm not sure this will actually work, but we'll find out.

### DIFF
--- a/docs/task-schema.md
+++ b/docs/task-schema.md
@@ -1,0 +1,15 @@
+---
+layout:       default
+class:        html
+docson:       true
+marked:       true
+ejs:          true
+superagent:   true
+docref:       true
+---
+
+# Task Definition
+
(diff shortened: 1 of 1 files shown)


djmitche:
Actually, it looks like tc-lib-docs supports this with a "schemas" key in its options, just not labeled in README.

djmitche:
Oh I see, those do not automatically generate top-level links.

djmitche:
(https://bugzilla.mozilla.org/show_bug.cgi?id=1326545 for adding that feature)

PR #138, opened 2017-01-13
(not ready for merging) Make task schema support arbitrary links

wlach:
This will be useful for https://bugzilla.mozilla.org/show_bug.cgi?id=1204490

### DIFF
--- a/schemas/task.yml
+++ b/schemas/task.yml
@@ -176,6 +176,31 @@ properties:
         type:           string
         format:         uri
         maxLength:      4096
+      links:
+        title:          "Links"
+        description: |
+          List of arbitrary links associated with the task. You could use this
+          for any purpose, one example might be a pointer back to a ci visualization
+          tool.
+        type:           array
+        items:
+          type:       object
+          title:      "Link pairs"
+          description: List of title/value link pairs
+          properties:
+            name:
+              description: Human readable description of where link goes
+              type:        string
+              maxLength:   255
+            url:
+              description: URL of link
+              type:        string
+              format:      uri
+              maxLength:   4096
+          required:
+            - name
+            - url
+          additionalProperties: false
     additionalProperties: false
     required:
       - name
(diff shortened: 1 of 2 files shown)


wlach:
@jonasfj @djmitche Here's a first stab at updating the schema. Do you have any suggestions on how to test my changes?

schemas/task.yml
```diff
@@ -176,6 +176,31 @@ properties:
...
+        type:           array
+        items:
+          type:       object
+          title:      "Link pairs"
+          description: List of title/value link pairs
+          properties:
+            name:
+              description: Human readable description of where link goes
+              type:        string
+              maxLength:   255
```
djmitche: why does this need a maximum length?

schemas/task.yml
```diff
@@ -176,6 +176,31 @@ properties:
...
+            url:
+              description: URL of link
+              type:        string
+              format:      uri
+              maxLength:   4096
```
djmitche: same..

schemas/task.yml
wlach: Just following the convention elsewhere in this file

jonasfj: maximum lengths are good... they prevent people from doing bad things... Notably, some crazy person like me might one day try to include a data url... which no longer metadata and definitely would be an abuse :)

jonasfj:
For the record, this needs a few tweaks in the code (I suspect), but otherwise the schema seems reasonable.

wlach:
Added some links to the create task test.

jonasfj:
One note:
The current `task.description` property is already declared to be markdown.
Hence, if these links are strictly for human consumption it might be just as fine to include a list of bulletpoints with links in the `task.description` property.

I'm just thinking that `task.description` would allow for nested lists, categories, header lines, etc. It's intended to be a human readable markdown text. Just a thought. At the end of the day, it might nicer to just have the links in the description, where other formatting options is available too.

wlach:
Yeah, I think we're going to go with the description approach for linking back to Treeherder, so closing this PR for now. We can always resurrect it if we decide we want a forest of links for tasks for some other reason. :)

PR #142, opened 2017-01-23
use res.reportError instead of res.status

djmitche:
This will give consistently-formatted API errors.

### DIFF
--- a/src/artifacts.js
+++ b/src/artifacts.js
@@ -100,13 +100,9 @@ api.declare({
   var past = new Date();
   past.setMinutes(past.getMinutes() - 15);
   if (expires.getTime() < past.getTime()) {
-    return res.status(400).json({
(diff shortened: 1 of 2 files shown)


djmitche:
I can't get the tests to run locally -- I suspect `user-config-example.yml`isn't correct?  So if this doesn't pass in travis, I may need your help.

djmitche:
OK, Travis seemed to run this correctly, and I got the tests running locally too.  Updated version should pass.

djmitche:
I have a taskwarrior item filed to find any stragglers in other services.  I suspect those will appear in hooks and secrets, where I seem to have made a habit of copying bad examples from other services!

jonasfj:
I'm sure index, auth, etc. has similar legacy stuff...

PR #144, opened 2017-01-26
initial work on blob storage type

jhford:
Just a quick review, feel free to ignore.  I'd like to do a more iterative approach to this project than a lot of other bigger projects which we do.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -9,25 +9,13 @@
       "optional": true
     },
     "accepts": {
-      "version": "1.1.4",
-      "from": "accepts@>=1.1.0 <1.2.0",
-      "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.1.4.tgz",
-      "dependencies": {
-        "mime-db": {
-          "version": "1.12.0",
-          "from": "mime-db@>=1.12.0 <1.13.0",
-          "resolved": "https://registry.npmjs.org/mime-db/-/mime-db-1.12.0.tgz"
-        },
-        "mime-types": {
-          "version": "2.0.14",
-          "from": "mime-types@>=2.0.4 <2.1.0",
-          "resolved": "https://registry.npmjs.org/mime-types/-/mime-types-2.0.14.tgz"
-        }
-      }
+      "version": "1.3.3",
+      "from": "accepts@>=1.3.3 <1.4.0",
+      "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.3.tgz"
     },
     "acorn": {
       "version": "4.0.4",
-      "from": "acorn@4.0.4",
+      "from": "acorn@>=4.0.1 <5.0.0",
       "resolved": "https://registry.npmjs.org/acorn/-/acorn-4.0.4.tgz"
     },
     "acorn-jsx": {
@@ -49,7 +37,7 @@
     },
     "ajv": {
       "version": "4.11.2",
-      "from": "ajv@>=4.11.2 <5.0.0",
+      "from": "ajv@>=4.8.2 <5.0.0",
       "resolved": "https://registry.npmjs.org/ajv/-/ajv-4.11.2.tgz"
     },
     "ajv-keywords": {
@@ -58,9 +46,9 @@
       "resolved": "https://registry.npmjs.org/ajv-keywords/-/ajv-keywords-1.5.1.tgz"
     },
     "amqplib": {
-      "version": "0.5.1",
-      "from": "amqplib@>=0.5.1 <0.6.0",
-      "resolved": "https://registry.npmjs.org/amqplib/-/amqplib-0.5.1.tgz",
+      "version": "0.4.2",
+      "from": "amqplib@>=0.4.0 <0.5.0",
+      "resolved": "https://registry.npmjs.org/amqplib/-/amqplib-0.4.2.tgz",
       "dependencies": {
         "isarray": {
           "version": "0.0.1",
@@ -100,9 +88,9 @@
       "resolved": "https://registry.npmjs.org/app-root-dir/-/app-root-dir-1.0.2.tgz"
     },
     "aproba": {
-      "version": "1.1.1",
+      "version": "1.0.4",
       "from": "aproba@>=1.0.3 <2.0.0",
-      "resolved": "https://registry.npmjs.org/aproba/-/aproba-1.1.1.tgz",
+      "resolved": "https://registry.npmjs.org/aproba/-/aproba-1.0.4.tgz",
       "optional": true
     },
     "are-we-there-yet": {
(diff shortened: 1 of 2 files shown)


src/data.js
```diff
@@ -263,45 +263,154 @@ let Artifact = Entity.configure({
...
+    'blobStore',
+    'privateBucket',
+    'publicBucket',
+    'monitor',
+  ],
+});
+
+/** Entity for tracking artifacts.
+ * Version 2 adds 'present' field */
+let Artifact = Entity.configure({
```
jonasfj: Won't work... You have to do the chaining

jonasfj: You can do `Artifact = Artifact.configure({version: 2`, but just use the format we had before...

Also you can't do `let Artifact` twice :)

src/data.js
```diff
@@ -263,45 +263,154 @@ let Artifact = Entity.configure({
...
-     *   path:          Path to blob within container
+     * storageType: blob
+     *   provider: which provider stores this artifact (e.g. s3)
+     *   region: which region this artifact is in (e.g. us-west-2)
+     *   bucket: name of the 'folder' the artifact is stored in (e.g.
+     *     taskcluster-public-artifacts)
+     *   key: name of the 'file' in the storage (e.g. $taskid/$runid/$name)
+     *   length: As per the HTTP Content-Length header
+     *   md5: hex encoded MD5 digest of content
+     *   sha256: hex encoded SHA256 digest of content
```
jonasfj: Couldn't we allign this docs as in the table format below?
Also intend text under NOTE:

src/data.js
```diff
@@ -263,45 +263,154 @@ let Artifact = Entity.configure({
...
+    expires: Entity.types.Date,
+    present: Entity.types.Number, // 0 or 1 to represent whether the artifact is there
+  },
+  migrate: function (item) {
+    // Add the present flag.  We will set this value to 1 for all cases because
+    // they are either stored in this table and can be worked on atomically so
+    // are always present if the entity can be loaded or because they are the
+    // deprecated s3 or azure storageType which will never have checking for
+    // presence
+    return {
```
jonasfj: No need to clone the object... just modify and return is legal:
```js
item.present = 1;
return item;
```

Or if you want to clone:
```js
return _.default({}, item, {present: 1});
```

src/data.js
```diff
@@ -263,45 +263,154 @@ let Artifact = Entity.configure({
...
 Artifact.prototype.json = function() {
   return {
-    storageType:      this.storageType,
-    name:             this.name,
-    expires:          this.expires.toJSON(),
-    contentType:      this.contentType,
+    storageType: this.storageType,
+    name: this.name,
+    expires: this.expires.toJSON(),
+    contentType: this.contentType,
```
jonasfj: This indentation style I understand, as it minimizes diffs... and doesn't hurt readability too much, but why do the same thing with the comments?

src/data.js
```diff
@@ -263,45 +263,154 @@ let Artifact = Entity.configure({
...
+/**
+ * A light wrapper around the create function that ensures some basic things
+ * are correct about the blobs, as well as forcing the present property to be 1
+ * for non-blob storageTypes and converting the truthiness of the present
+ * property to 0 or 1
+ *
+ * NOTE: I have included code here in comments which should be uncommented
+ * when we wish to bar creation of new s3 or azure entities
+ */
+Artifact.create = function(properties, overwriteIfExists /*, allowDeprecated=false*/) {
```
jonasfj: I suggest that we don't do this...

If you really want validation I propose that we add a new type to azure-entities.
Instead of `Entity.types.JSON` we could do: `Entity.types.Schema({type: object, ...})`, and then let the type wrap a JSON schema to ensure validation...

src/data.js
```diff
@@ -310,23 +419,39 @@ Artifact.prototype.json = function() {
...
  * artifact.
  */
-Artifact.prototype.remove = function(ignoreError) {
+Artifact.prototype.remove = async function(ignoreError) {
   // Promise that deleted underlying artifact, and keep reference to context
   var deleted = Promise.resolve();
 
+  // Shorthand because we refer to this property so many times.  Saves
+  // prototypechain lookups and linespace
+  let dtl = this.details;
```
jonasfj: Could at least make it `details`, if not how about we destructure it instead?

src/data.js
jhford: heh, i haven't actually run this code yet :)

src/data.js
jhford: In the past you said that I shouldn't do that, which is why I created a new object.  I'll happily do that

src/data.js
jhford: Because i don't have whatever thing it is you do that automatically does that fancy indentation.  I use the vim auto-format stuff which just gives you nice <=80 char lines, which doesn't understand the indentation format where things line up.

I can change comments though

src/data.js
jhford: There are things that I'd like to do which cannot be done with JSON Schema, e.g. the deprecation for which the code is currently commented out.  For a lot of this, we can rely on the API's JSON Schema.

Why is it that you suggest not doing this?  I can't see any issues with wrapping the `create` method with a couple checks, especially as we have a similar wrapper for the `remove` method.

src/data.js
jhford: Destructuring would make it more complicated, imo, and would require destructuring in three different places (once for blob, once for s3 and once for azure).  I'll rename to details.

src/data.js
jonasfj: > Because i don't have whatever thing it is you do that automatically does that fancy indentation

OCD? :)

src/data.js
jonasfj: it's just ugly an unnecessary, if we want assertions like this I propose we express them in a declarative schema.

Conditions you expressed can certainly be implemented using JSON schema.

---
I'm suggesting this to keep things simple. IMO we don't need these checks. But I see how they are super nice to have, so if we want them I suggest we do it declaratively.

jonasfj: @jhford,
Can you see how you can use the:
```json
Entity.types.Schema({
  type: 'object',
  properties: {
    md5: {type: 'string', pattern: '^[a-fA-F0-9]{32}$'},
    sha256: {type: 'string', pattern: '^[a-fA-F0-9]{64}$'},
  },
});
```

You can even use `anyOf` and do more complication validation if you wish...

src/data.js
```diff
@@ -263,45 +263,154 @@ let Artifact = Entity.configure({
...
+  // 1 because all other artifacts are either atomically stored in Azure Tables
+  // or are stored in deprecated storageTypes which will never have a check for
+  // present status.  For blob storage items, we'll have as a convenience the
+  // ability to support specifying the present field as either a number or a
+  // boolean.  We will only accept primatives number or boolean
+  if (props.storageType === 'blob' && typeof props.present === 'number') {
+    assert(props.present === 0 || props.present === 1, 'Present must be 0 or 1');
+    props.present = props.present
+  } else if (props.storageType === 'blob' && typeof props.present === 'boolean') {
+    props.present = props.present ? 1 : 0;
```
jonasfj: Let's not do this... This won't work in `.modify`...
I agree that it's not pretty to use `1` and `0` as booleans. So if you don't want to do that, just add a type to `azure-entities`...
No reason we can't invent `Entity.types.Bool`. Azure does support booleans `Edm.Boolean`, they are just encoded as booleans.
We would need a test for the boolean type to make sure that filtering works, but other than that it should be pretty straight forward to add a new type to `azure-entities`

PR #146, opened 2017-02-09
Extra asserts

jonasfj:


src/api.js
```diff
@@ -1286,6 +1285,18 @@ api.declare({
...
+// for each, instead we create a new promise every 2s and reuse that.
+let _lastTime = 0;
+let _sleeping = null;
+let sleep20Seconds = () => {
+  let time = Date.now();
+  if (time - _lastTime > 2000) {
+    _sleeping = new Promise(accept => setTimeout(accept, 20 * 1000));
+  }
+  return _sleeping;
+};
```
jonasfj: If this ugly hack works... I try to encapsulate it in a different file... make it less ugly :)

PR #147, opened 2017-02-14
fix spelling errors

djmitche:
As found in taskcluster/taskcluster-cli#100


jonasfj:
@djmitche for future reference please push to the taskcluster/taskcluster-queue repo, or tests won't run...

PR #149, opened 2017-02-15
Found a bug with timedout requests not being removed, not sure how mu…

jonasfj:
…ch it explains.

### DIFF
--- a/src/workclaimer.js
+++ b/src/workclaimer.js
@@ -2,6 +2,7 @@ let assert      = require('assert');
 let _           = require('lodash');
 let events      = require('events');
 let taskcluster = require('taskcluster-client');
+let Promise     = require('promise');
 
 /**
  * HintPoller polls for hints for pending tasks.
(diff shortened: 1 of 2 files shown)


src/workclaimer.js
```diff
@@ -25,22 +26,22 @@ class HintPoller {
...
-      request = {resolve, reject, count};
+    return new Promise((resolve, reject) => {
+      // Make a request for count tasks
+      let request = {resolve, reject, count};
+      this.requests.push(request);
+
+      // Remove request if aborted
+      aborted.then(() => {
+        // Remove request from requests, but modifying the requests array
+        _.pull(this.requests, request);
```
jonasfj: We used `_.remove` before, but that didn't work... I'm not sure what negative side-effects this would have... Could have implications on memory...

src/workclaimer.js
```diff
@@ -25,22 +26,22 @@ class HintPoller {
...
+        // Resolve request empty array
+        request.resolve([]);
+      }).catch(err => process.nextTick(() => this.parent.emit('error', err)));
```
jonasfj: @imbstack, do you agree that this would likely show up in sentry, I'm like 30%  sure I've done that right here and below...

PR #151, opened 2017-03-07
Limit task.routes to 10

jonasfj:
This is only on the create task request, not on the response since that would break what we currently have stored in the queue.

### DIFF
--- a/schemas/create-task-request.yml
+++ b/schemas/create-task-request.yml
@@ -91,7 +91,7 @@ properties:
       type:         string
       maxLength:    249
       minLength:    1
-    maxItems:       64
+    maxItems:       10 # 4kb frame limit, 10 * 255 = 2550 that leaves a few bytes for other fields
     uniqueItems:    true
   priority:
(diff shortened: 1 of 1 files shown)


jonasfj:
Yeah, it'll break things for Callek...

But afaik he'll be willing to fix that when we've fixed:
https://bugzilla.mozilla.org/show_bug.cgi?id=1333255

So yeah, maybe we wait a bit before we land this :)
And make sure callek has had a chance to uplift to various branches.

jonasfj:
@Callek, do you have things that would break now if we limit to 10 routes again... This will improve stability.

djmitche:
I added a reminder for myself to remind y'all to land this :)

djmitche:
@Callek should we proceed?

Callek:
I say yes, -- There's a (small) chance we'll need to backout temporarily if some releasy tree isn't ready, but I don't think we need to worry at present.

djmitche:
@jonasfj want to land this?

PR #152, opened 2017-03-07
Lets lock node to 7.6 for now

jonasfj:


imbstack:
I'll get this in my yarnification stuff.

PR #154, opened 2017-03-13
Use the same babel-preset that we're using elsewhere

jhford:
We're using a really ancient version of the babel-preset.  I'm not sure why, but maybe we should update it to the latest.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "private": true,
   "scripts": {
-    "compile": "babel-compile -c tc-rules/babel src:lib test:.test",
+    "compile": "babel-compile -p taskcluster src:lib test:.test",
     "test": "./test/runtests.sh",
     "pretest": "npm run compile",
     "install": "npm run compile"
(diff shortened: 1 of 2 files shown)


jonasfj:
Log from crash after we deployed this:
https://pastebin.mozilla.org/8981904

I have no idea why, it makes zero sense,, I can't reproduce locally... Or on travis.

jonasfj:
On heroku the build was:
https://pastebin.mozilla.org/8981905

Still no idea...

jhford:
that's usually because you're trying to run node 0.12 with a `node_modules` built with >= 4 or the other way around.  It's because 4 broke the buffer api.  Maybe we can force all the heroku caching to drop itself?

imbstack:
I believe this failed because of issues with yarn+heroku discussed in https://github.com/heroku/heroku-buildpack-nodejs/issues/380#issue-211999514

jhford:
@jonasfj, @imbstack do you think we can try to land this again?

imbstack:
I think this is landed. We didn't revert it when prod broke since this didn't cause the issue.

PR #155, opened 2017-03-13
[DON'T MERGE] New Blob Artifact type

jhford:
This PR is so that I can run unit tests because they don't run locally


src/artifacts.js
```diff
@@ -4,8 +4,41 @@ let assert  = require('assert');
...
 let API     = require('taskcluster-lib-api');
 let urllib  = require('url');
+let crypto  = require('crypto');
 let api     = require('./api');
 
+/**
+ * Encode a storage key for storing in the blob storage type backends We do
+ * this to get better partitioning in S3 which gives us better performance.
+ *
+ * Basically `${md5hash(taskId + runId + name).slice(0,7)}/${taskId}/${runId}/${name}`
```
jonasfj: @jhford, given how random `taskId` already is... is there any value to this?

Compared to the value of being able to access artifact even if the azure table storage is lost...
I don't say I want to plan for that happen, it would make me VERY sad, and would be super super bad. But in terms of simplicity it seems to have some value to use keys `${taskId}/${runId}/${name}`.

And there is no win by putting md5 in-front, since `taskId` is random base64 encoded. It has more entropy than hexadecimal md5. Ie. uses more characters...

Note: the odds that splitting artifacts from a single task to multiple s3 partitions has any value is very small..

src/artifacts.js
```diff
@@ -176,7 +220,31 @@ api.declare({
...
+      details.region = 'us-east-1'; // MAKECONFIG
+      if (input.encoding) {
+        details.encoding = input.encoding;
+      }
+      if (isPublic) {
+        details.bucket = 'taskcluster-blobs'; // MAKECONFIG
+      } else {
+        details.bucket = 'taskcluster-private-blobs'; // MAKECONFIG
+      }
+
```
jonasfj: Shouldn't we just initiate the multipart upload up here?

jonasfj: That way we avoid the `.modify` operation below...

src/artifacts.js
jhford: We could always reconstruct it by listing the objects in the buckets and then trimming off the first path, but I think you're right that random taskIds should be enough randomness for the partitioning impact.  I'd just forgotten about that detail :)

src/artifacts.js
jhford: Except that if we initiate the upload and the Entity creation fails, we have an upload floating around.  I thought about doing this, but I thought it was safer to not do it that way.  I know that the bucket policy should clean up those uploads, but it seems so easy to avoid that situation in the first place.

src/artifacts.js
```diff
@@ -254,23 +325,85 @@ api.declare({
         });
     }
 
+    if (artifact.storageType === 'blob') {
```
jonasfj: We need to modify the `_.isEqual` above... so that it ignores the `uploadId`...

If we have two concurrent calls to `createArtifact` and both of them creates initiates upload and creates the artifact entity we just drop the last one of the `uploadId`s...

Note: I find this rather extremely unlikely happen... It could happen if the `createArtifact` call takes more than 30s and the worker, hence, retries the request. So mostly I care that we handle the case in a consistent manner. Cancelling the `uploadId` would be nice, but how often is going to happen that we run into this... I'm guessing we can ignore cancellation and just let the uploadId time out in that instance...

src/artifacts.js
```diff
@@ -176,7 +220,31 @@ api.declare({
   var isPublic = /^public\//.test(name);
   var details  = {};
   switch (storageType) {
```
jonasfj: @jhford, I feel I must apologizes for the horrors that is the artifact API end-point implementation... It's so messy...

If you 110% fired up feel free to refactor all of this into different functions... perhaps it would have been smartest to just start the this handler with:
```js
switch (storageType) {
  case 'blob':
    return createAritfactBlob(req, res);
  case ...
}
```

and have separate methods for each artifact type... I know it would imply some code duplication, but right now we have sprinkled special-cases all over the place :)

src/artifacts.js
jonasfj: Yeah, going back and forth over this... it seems that no matter what we do we'll have a race where we risk creating an `uploadId` too much. Right?

Ie. we could have two racing later down in the `.modify` section... in fact it's probaby problematic to just blindly overwrite it... concurrency in the morning is hard. So only 92% sure at this point :)

Regardless, we're not going to race very often, so really however we resolve it doesn't matter... because it'll only happen every odd moon (is that an expression?).

src/artifacts.js
```diff
@@ -254,23 +325,85 @@ api.declare({
...
+            key: artifact.details.key,
+            sha256: artifact.details.sha256,
+            size: artifact.details.size,
+            metadata: {taskId, runId, name},
+            tags: {taskId, runId, name},
+            contentType: artifact.contentType,
+            contentEncoding: artifact.details.encoding || undefined,
+          });
+          await artifact.modify(_artifact => {
+            _artifact.details.uploadId = uploadId;
```
jonasfj: We should do:
```js
await artifact.modify(artifact => {
  if (!artifact.details.uploadId) {
    artifact.details.uploadId = uploadId;
  }
});
uploadId = artifact.details.uploadId;
```

That way if we're racing it's always the first one that wins...

We could add:
```js
if (artifact.details.uploadId != upload) {
  await s3.cancelMultiPartUpload(uploadId);
  uploadId = artifact.details.uploadId;
}
```
But the cancel part is just nice...

Note: I think all of this should be part of the create entity phase... and that we then handle this cancel uploadId part if there is a create conflict...

jonasfj: This is a a bit complicated... Particularly, handling the `_.isEqual` part of the create entity flow...

But if we're lucky it might just turn up somewhat elegant in the end :)
I mean cancelling the uploadId if there is a conflict might not be a hard I thought...

src/artifacts.js
jhford: I'll remove the MD5 stuff

src/artifacts.js
jhford: I think the advantages of not doing a `.modify()` are worth it.  Let's do that.

jhford:
Hi @jonasfj please take a look at the changes that I've made today.  It uses the control flow that you suggested, basically we'll now create a new `uploadId` each time we create an `Artifact`.  If there's a collision and the only thing that's differing between the stored and new entity, we'll cancel the freshly created upload and use the existing one.  This should allow us to be able to regenerate request lists.

src/artifacts.js
```diff
@@ -175,8 +217,46 @@ api.declare({
...
+        details.encoding = input.encoding;
+      }
+
+      if (isPublic) {
+        details.bucket = 'taskcluster-blobs'; // MAKECONFIG
+      } else {
+        details.bucket = 'taskcluster-private-blobs'; // MAKECONFIG
+      }
+
+      details.key = encodeBlobKey(taskId, runId, name);
```
jonasfj: please drop this see previous discussion.

jhford: whether we do the MD5 thing or not, I still think that having a single place where we map taskId, runId and name to an S3 key is worthwhile.

jonasfj: I don't care... But storing it seems overkill... It's just a function mapping to string... Joining with slash...

src/artifacts.js
```diff
@@ -175,8 +217,46 @@ api.declare({
   // Construct details for different storage types
   var isPublic = /^public\//.test(name);
   var details  = {};
+  let uploadId;
   switch (storageType) {
+    case 'blob':
+      // Generate the details that we'll be using.
+      details.size = input.size;
```
jhford: We can do that as an added validation, but I would like the creator to tell us what it thought the overall size is.  This is about verification and validation, I'd rather not compute things we receive from the generating worker where we can easily pass in the value the worker thought it was

src/artifacts.js
```diff
@@ -175,8 +217,46 @@ api.declare({
...
+      details.sha256 = input.sha256;
+      if (input.parts) {
```
jhford: If we `createArtifact` a second time with a different set of parts then it's an idempotency error if we return the same upload id for a different set of parts.

Why would we care to transform the parts when we can just store the original one and always work off it?  Maybe we can delete the parts list during `completeArtifact`, but I'd rather that we keep that info around for later examination.

src/artifacts.js
```diff
@@ -175,8 +217,46 @@ api.declare({
...
+
+      if (isPublic) {
+        details.bucket = 'taskcluster-blobs'; // MAKECONFIG
+      } else {
+        details.bucket = 'taskcluster-private-blobs'; // MAKECONFIG
+      }
+
+      details.key = encodeBlobKey(taskId, runId, name);
+      if (input.parts) {
+        uploadId = await this.s3Controller.initiateMultipartUpload({
```
jhford: We are choosing to understand the S3 Rest API by doing this remote signing of requests in the first place.  I've chosen to implement this as a library without the `aws-sdk` library but rather the `aws4` signing library.  Since we need to understand how to generate the general form of an S3 rest API request, and we also need to understand how to run a request of that form, implementing the init/complete/abort calls is really trivial.  There's a lot of tests to cover these api calls.

Having this as a self-contained library also means that we can import the library already knowing that all of these methods work, and more importantly work together.  We can have testing in the queue be about testing how the queue works, and not about the low level details about how S3 uploading works.

This library also is written in a very general way which allows others in the open source community to benefit from the work we're doing here, without having to import the entire `aws-sdk` library or figure out how to use them together.

The S3 Rest API is very well documented and has a lot of information on how to use it.  It's not like we're using some sort of undocumented private API.  The `aws-sdk` library is merely a translator from Javascript to identical REST Api that I'm calling directly.  This library is not a general purpose AWS or even a general purpose S3 library.  It's a library which knows how to do uploads to S3 with a remote signing server.

If you have specific technical concerns, I am happy to address them or rethink the problem as I have done many, many times through this project.

Furthermore, this is for a new artifact type which does not have any production consumers yet.  If we land it and in testing find that this library is fundamentally unsuited to the task, I'm happy to rethink it, but this is the approach that I'm taking until there's real evidence to the contrary.

As for maintenance, I think it's highly improbable that AWS will break the S3 apis in a non-compatible way.  If they did, however, we'd be just as broken since we depend on 100% of the core functionality of the `remote-signed-s3` library working.  That doesn't change whether we call init/complete/abort through `remote-signed-s3` or `aws-sdk`.

src/artifacts.js
```diff
@@ -261,16 +378,55 @@ api.declare({
...
+          contentType: artifact.contentType,
+          contentEncoding: artifact.details.encoding || undefined,
+        });
+        requests = [singlePartRequest];
+      }
+      res.reply({
+        storageType: 'blob',
+        contentType: artifact.contentType,
+        contentLength: artifact.size,
+        contentSHA256: artifact.sha256,
```
jhford: I don't think it hurts to include them

src/artifacts.js
```diff
@@ -261,16 +378,55 @@ api.declare({
...
+        requests: requests,
```
jhford: nope, because requests is a list of requests.  if we were returning information about parts, and that list was a list of parts, then we'd call it parts.

src/artifacts.js
```diff
@@ -415,6 +626,99 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+    if (artifacts.details.parts) {
+      etag = await this.completeMultipartUpload({
+        bucket: artifact.details.bucket,
+        key: artifact.details.key,
+        etags: input.etags,
+        uploadId: artifacts.details.uploadId,
+        tags: {taskId, runId, name},
+      })
+    } else {
+      etag = input.etag;
```
jhford: yes, that's something that I intend to do, as I mentioned in the comment 2 lines below

```diff
+    // TODO: Do a quick HEAD of the resource (maybe) to double check that the
+    // expected x-amz-meta-content-sha256 value matches and that the etag we
+    // either received or generated by completing a multipart upload matches
+    // the expected value
```

src/data.js
```diff
@@ -290,6 +290,30 @@ let Artifact = Entity.configure({
...
+    /** 
+     * Present is a number field which represents an artifact being present and
+     * the upload being completed.  The handling logic for the artifact's
+     * storage type will fully define what that means for a given storage type.
+     *
+     * Because there are no boolean fields in Azure-Entities which allow for
+     * fast scanning based on properties (we could use JSON), we're using a
+     * number in place of a boolean.  0 means false, 1 means present
+     */
+    present:        Entity.types.Number,
```
jhford: Let's not block this project on implementing Boolean in azure-entities.  I'm happy to take a look at it in the future, but storing 1 or 0 in a number field should be fine for now.

src/artifacts.js
jonasfj: for size and performance, we're probably best off hashing the parts... and refer to the request body when using it... (ofcourse checking the hash)

PR #157, opened 2017-03-13
Proposed schema for create artifact request

jonasfj:
Note: if there is just one part, then you duplicate `sha256` and `md5` in `parts[0].sha256`  and `parts[0].md5`, but this seems acceptable.

It certainly reduces the number of cases we have to handle :)

### DIFF
--- a/schemas/post-artifact-request.yml
+++ b/schemas/post-artifact-request.yml
@@ -3,6 +3,49 @@ title:              "Post Artifact Request"
 description: "Request a authorization to put and artifact or posting of a URL as an artifact. Note that the `storageType` property is referenced in the response as well."
 type:               object
 oneOf:
+  - title:          "Blob Artifact Request"
+    description: |
+      Request for to upload a generic blob artifact.
+    type:           object
+    properties:
+      storageType:
+        description: |
+          Artifact storage type, in this case `'blob'`
+        type:       string
+        enum:
+          - blob
+      contentType:
+        description: |
+          Artifact mime-type, when uploading artifact to the signed
+          `PUT` URL returned from this request this must given with the
+           `ContentType` header. Please, provide correct mime-type,
+           this make tooling a lot easier, specifically,
+           always using `application/json` for JSON artifacts.
+        type:       string
+        maxLength:  255
+      sha256:       {type: string, pattern: '^[0-9a-f]{64}$'}
+      md5:          {type: string, pattern: '^[0-9a-f]{32}$'}
+      parts:
+        type:       array
+        items:
+          type:     object
+          properties:
+            sha256:       {type: string, pattern: '^[0-9a-f]{64}$'}
+            md5:          {type: string, pattern: '^[0-9a-f]{32}$'}
+            size:         {type: integer, minimum: 0}            
+          additionalProperties: false
+          required:
+            - sha256
+            - md5
+            - size
+        minItems: 1
+    additionalProperties: false
+    required:
+      - storageType
+      - contentType
+      - sha256
+      - md5
+      - parts
   - title:          "S3 Artifact Request"
     description: |
       Request for a signed PUT URL that will allow you to upload an artifact
--- a/schemas/post-artifact-response.yml
+++ b/schemas/post-artifact-response.yml
@@ -47,9 +47,7 @@ oneOf:
     additionalProperties: false
     required:
       - storageType
-      - putUrl
-      - expires
-      - contentType
+      - parts
   - title:          "S3 Artifact Response"
     description: |
       Response to a request for a signed PUT URL that will allow you to
--- a/schemas/post-artifact-response.yml
+++ b/schemas/post-artifact-response.yml
@@ -5,6 +5,51 @@ description: |
   Note that the `storageType` property is referenced in the request as well.
 type:               object
 oneOf:
+  - title:          "Blob Artifact Response"
+    description: |
+      Response to a request for creating a generic blob artifact.
+    type:           object
+    properties:
+      storageType:
+        description: |
+          Artifact storage type, in this case `'blob'`
+        type:       string
+        enum:
+          - blob
+      parts:
+        type:       array
+        items:
+          type:     object
+          properties:
+            method:
+              type: string
+              enum:
+                - put
+                - post
+                - delete
+                - patch
+                - get
+            url:
+              type: string
+              format: uri            
+            headers:
+              type: object
+              additionalProperties:
+                type: string
+            size: {type: integer, mimimum: 0}
+          additionalProperties: false
+          required:
+            - method
+            - url
+            - headers
+            - size
+        minItems:   1
+    additionalProperties: false
+    required:
+      - storageType
+      - putUrl
+      - expires
+      - contentType
   - title:          "S3 Artifact Response"
     description: |
       Response to a request for a signed PUT URL that will allow you to
--- a/src/artifacts.js
+++ b/src/artifacts.js
@@ -35,19 +35,30 @@ api.declare({
     'intermediate artifacts from data processing applications, as the',
     'artifacts can be set to expire a few days later.',
     '',
-    'We currently support 4 different `storageType`s, each storage type have',
+    'We currently support 3 different `storageType`s, each storage type have',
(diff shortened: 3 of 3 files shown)


jonasfj:
@jhford, also note I propose that `blob` artifacts derive their expiration from `task.expiration` instead of having it specified... We talked about that earlier too.

schemas/post-artifact-request.yml
```diff
@@ -3,6 +3,49 @@ title:              "Post Artifact Request"
...
+            - md5
+            - size
+        minItems: 1
+    additionalProperties: false
+    required:
+      - storageType
+      - contentType
+      - sha256
+      - md5
+      - parts
```
jhford: As I've mentioned before, there's a difference between a single part upload (`PUT` for object) and a multipart upload that has only one part.  I would rather signify that difference by the presence of the parts list.  We'd have to have an extra check that the single part's sha256 matches the file's sha256.  That's not possible in JSON Schema, at least not easily as best as I can tell.  Further, it doesn't match the intent.  There might be some practical reason for using multipart uploads for single part uploads in the future.  I'd rather be explicit about this, especially as figuring out the basics of the upload is a client responsibility.

How about we compromise and have the schema say that presence of the parts list implies multipart upload, and that if specified we ensure that there are two or more items in the list.

schemas/post-artifact-request.yml
```diff
@@ -3,6 +3,49 @@ title:              "Post Artifact Request"
...
+        description: |
+          Artifact mime-type, when uploading artifact to the signed
+          `PUT` URL returned from this request this must given with the
+           `ContentType` header. Please, provide correct mime-type,
+           this make tooling a lot easier, specifically,
+           always using `application/json` for JSON artifacts.
+        type:       string
+        maxLength:  255
+      sha256:       {type: string, pattern: '^[0-9a-f]{64}$'}
+      md5:          {type: string, pattern: '^[0-9a-f]{32}$'}
```
jhford: let's skip md5.  Because of multipart uploads, we cannot rely on the `Content-MD5` header.  That header was the only value of using the obsolete MD5 hashing algorithm.

schemas/post-artifact-request.yml
```diff
@@ -3,6 +3,49 @@ title:              "Post Artifact Request"
 description: "Request a authorization to put and artifact or posting of a URL as an artifact. Note that the `storageType` property is referenced in the response as well."
 type:               object
 oneOf:
+  - title:          "Blob Artifact Request"
+    description: |
+      Request for to upload a generic blob artifact.
+    type:           object
+    properties:
```
jhford: We're missing the overall size of the file in this object.

jonasfj: We don't need it, we can just sum up the parts :)

schemas/post-artifact-request.yml
jonasfj: Okay, I agree... 

Randomly, should we consider adding another algorithm while we're at it?
I just saw this on review of hash algorithms for git by some crypto experts:
https://public-inbox.org/git/91a34c5b-7844-3db2-cf29-411df5bcf886@noekeon.org/

I'm quite happy with sha2-256, but adding SHAKE128 or PH128 on top of that might not be entirely crazy. Just if we are supposed to think forward a few years.

schemas/post-artifact-request.yml
jonasfj: Summary, we have two options:
A) Require that `parts` is always present, use `parts.length == 1` for single-part uploads, making single part uploads opaque to the REST API consumer.
B) Make `parts` optional, and require `parts.length >= 2` if present, have single-part upload be an explicit special case as far as the REST API consumer is aware.

This discussion has some aspects of bike-shredding... In that either solution will work and choice is probably not super important. That said I've been asked to try an explain more rather than just make unfounded statements. Also while the difference between (A) and (B) isn't huge or world changing in anyway, I suspect that they are results of differing underlying assumptions about the public API.

I see the difference in preference for the two options is the result of a difference in thinking. Depending on whether we think that:
i) Consumers will always use the REST API through a library we provide, or,
ii) Consumers will use the REST API as is documented in schema and references.  

If we assume (i), that consumers will always **use the REST API through a library** we've implemented, then:
1) Choice of (A) and (B) is utterly unimportant to the consumer.
2) Choice of (A) and (B) is of less importance to us too, because it's hidden by the library.
3) Being slightly more explicit as is the case in (B) is makes some sense, because we 100% control both sides of the REST API.

If we assume (ii), that consumers will **use  the REST API as documented** in references and schemas, then:
1) Choice of (A) reduces number of optional properties making @petemoore happier when we auto-generate golang and java client libraries. Optional properties are still better than `anyOf`s, but for static language support we really prefer determinism. Notice that the response would have to be an `anyOf`, which we then have to manually parse from `json.RawMessage` in golang.
2) Choice of (A) will make us happier in the future when we get around to doing a rust client. Same reasoning as in (1).
3) Choice of (B) requires consumers to implement two different code-paths for artifact upload depending on whether not there is one or more multiple parts. This in turn increases number of tests cases and risk of bugs. Option (A) allows the same code-path to be used for multi-part and single-part upload. In effect you would only write code-path that does multi-part upload and use that  with `N=1` if doing a single part upload. This would all be opaque to the API consumer.
 
**In conclusion** I absolutely see a slight preference for (B) if we accept (i) as the school of thought (naming that consumers will always use the REST API through an artifact library we've implemented). If, however, we expect consumers to be using the REST API directly, I see a decent preference for option (A). Notably reducing in code-paths and stronger typing for both `createArtifact` request and response as well as the `completeArtifact` request (where we are likely having a similar argument about whether or not to send `parts[{etag: ...}]` as payload even in the case of single-part upload.

-----
However, I don't buy into the feasibility of (i), which I think is the root of this discussion and why this is worth debating. If this wasn't rooted in different schools of thought it would be pretty much a bike-shredding exercise :)

So far we've design taskcluster as a set of loosely coupled components. With the interface between different services being a REST API. Ensuring that we don't lock ourselves into a specific technology or language. This has serves us very well. Uprooting that and requiring consumers of the API to use a specific library/language or set of languages, would be dramatic shift in design policy. Such a dramatic shift would in my opinion require a broad and wide consensus, both from within our team, but likely also from stakeholders outside out team.

While people on-board with this idea (i) isn't impossible it would be very hard. I for one would have to have to arguments reiterated many times. Even if I buy the arguments the first time I hear them, it would take long time to turn around such a fundamental core principal. It took me a long time to convince myself that task dependency scheduling belonged in the queue.

That said, the arguments for (i) are strong, namely that consumers repeated get things wrong when interacting with the artifact API. We haven't had huge issues people using the upload API wrong, mostly the API just didn't provide sufficient safety (rather than it being used incorrectly).
Nevertheless, the notion is correct that the upload API is complicated and mistakes such as failing to refresh signatures is easy to make.

So I would suggest that we do what we've done so far... We design a public REST API and offer client libraries for using it. We've had problems with people not retrying requests or failing to sign requests correctly, these have largely been alleviated by implementation of the taskcluster client libraries in node, go and python. However, using these libraries is not required.
I suggest we go the same route with this. We first and foremost design for a public REST API, with (ii) as the school of thought. But we build client libraries in popular languages and strongly encourage their adoption by consumers. If we make the right thing easy to do, then that's typically enough.

----
Okay, that was my long reasoning... Sorry, this go so crazy long.
Note: I'm well aware that the `createArtifact` api already contains `anyOf`s, but I would really like to avoid introducing another level of it, and most certainly would like to avoid context conditional properties for `completeArtifact`. I'm have some early stage ideas for improving our json-schema to golang code-gen, that will likely make some of this nicer, but two levels of `anyOf` makes things even harder :)
Note: optional properties is better than `anyOf`, but I very much want to avoid optional properties in return values. Generally, optional properties are really bad, if they don't just specify a default value (that's pretty much the only way we use the right now). If we really want to be explicit an different artifact kind might be better (but this also feels sad).

jonasfj: Okay, if the response format is the same I don't care anymore... :)

jhford:
this is being handled in the artifact api pr.

PR #158, opened 2017-03-16
Docs for reclaimTask

jonasfj:


jonasfj:
yes, this is another way to find out that your task has been cancelled..

But polling every 30s won't scale to 3-5k workers, well, it might, but please don't add that load :)

imbstack:
Ah makes sense. I've added my thoughts to https://bugzilla.mozilla.org/show_bug.cgi?id=1348121

PR #159, opened 2017-03-23
[DON'T MERGE] Blob artifacts (new PR)

jhford:
The old PR was starting to get a little busy.  I wanted to start a new one.

There's some things left still to do, including:

* ensure `.remove()` and `.expire()` or w/e it's called work with blobs
* implement a delete in remotely-signed-s3 to cleanup in the unit tests and for the above methods
* give the ETag header on the `getArtifact` redirect response.  We need this because we need to allow the client to be able to see what the ETag was supposed to have been at creation.  I can't think of a better way to pass this information on without either dropping the 303 redirect or creating an extra endpoint
* decide if we should check the ETags automatically with a HEAD on every getArtifact call.  Adds overhead but is automatic consistency checking and ensures we haven't changed the artifact out from underneath our view of the artifact table
* the `createBuckets` component I added to main.js should set the public and private access and cors polcies as we wish them to be
* I've removed the `aws-sdk-promise` library from the queue... It breaks the aws-sdk library and I really don't think we should have it.  I'm not sure if I made all the changes required, please double check
* I need to check that the `completeArtifact` message is sent -- hints appreciated, it's been a while since I've done this
* switch to the new boolean entity column type

Otherwise, this should be in a good state.  I should probably write a couple more tests around public vs. private handling, but it is working really well.


test/artifact_test.js
```diff
@@ -15,6 +15,15 @@ suite('Artifacts', function() {
   var {Netmask}     = require('netmask');
   var assume        = require('assume');
   var helper        = require('./helper');
+  var fs            = require('fs');
+  var crypto        = require('crypto');
+  var randbytes     = require('randbytes');
+  var tmp           = require('tmp');
+  var remoteS3      = require('remotely-signed-s3');
+  var qs            = require('querystring');
+  var urllib        = require('url');
```
jhford: Yes, this name is intentional.  There are already lines with `var url` and I don't want mess with those.

jonasfj: just saying other places we've often used `URL = require('url')` to avoid the problem with `url` variable...

Note, in golang I have the same issue, and have started prefering `u` instead of `url` as variable name.

Regardless, this is nits...

jonasfj:
> I need to check that the completeArtifact message is sent -- hints appreciated, it's been a while since I've done this

Take a look at existing tests... `ListenFor` starts listening:
https://github.com/taskcluster/taskcluster-queue/blob/8716db894eb0bbcbe3e4c1508040dce2e20d4966/test/artifact_test.js#L445-L447
`waitFor` waits for the message previously listened for:
https://github.com/taskcluster/taskcluster-queue/blob/8716db894eb0bbcbe3e4c1508040dce2e20d4966/test/artifact_test.js#L463

schemas/post-artifact-request.yml
```diff
@@ -3,6 +3,110 @@ title:              "Post Artifact Request"
...
+        type: string
+        maxLength: 255
+      sha256:
+        description: |
+          The complete SHA256 value of the entire artifact.  This must be the
+          SHA256 of the file which is to be uploaded.  For single part uploads,
+          the upload will fail if the SHA256 value of what is uploaded does not
+          match this value
+        type: string
+        pattern: "[a-fA-F0-9]{64}"
```
jonasfj: you need to anchor this regex, `"^[a-fA-F0-9]{64}$"`

schemas/post-artifact-request.yml
```diff
@@ -3,6 +3,110 @@ title:              "Post Artifact Request"
...
+          The number of bytes of the entire artifact.  This must be the number
+          of bytes in the file to be uploaded.  For single part uploads, the
+          upload will fail if the number of bytes uploaded does not match this
+          value.  A single part upload (e.g. no parts list) may be at most 5GB.
+          This limit is enforced in the code because it is not possible to
+          represent all of the restrictions in a json-schema.  A multipart
+          upload may be at most 5TB, with each part other than the last being
+          between 5MB and 5GB in size.
+        type: number
+        multipleOf: 1.0
```
jonasfj: No need, use `type: integer` instead... It's a bit weird that JSON schema supports both `number` and `integer` types, but when you have to do code-gen it's really awesome to have `type: integer`.

schemas/post-artifact-request.yml
```diff
@@ -3,6 +3,110 @@ title:              "Post Artifact Request"
...
+              maxLength: 64
+            size:
+              description: |
+                The number of bytes in this part.  Keep in mind for S3 that
+                all but the last part must be minimum 5MB and the maximum for
+                a single part is 5GB.  The overall size may not exceed 5TB
+              type: number
+              multipleOf: 1.0
+              minimum: 0
+            start:
```
jonasfj: It feels like this is implied as the sum of all the previous part? Also it feels like this is very uninteresting to the queue.

Also it could give people the impression that they can place order the parts as they wish, which they can't...

src/main.js
```diff
@@ -277,13 +279,79 @@ let load = loader({
...
+      let s3 = new aws.S3({
+        region: cfg.app.blobArtifactRegion,
+        accessKeyId: cfg.aws.accessKeyId,
+        secretAccessKey: cfg.aws.secretAccessKey,
+      });
+
+      // First, we'll create a public bucket
+      let publicConfig = {
+        Bucket: cfg.app.publicBlobArtifactBucket,
+        ACL: 'public-read',
```
jonasfj: we don't not use ACLs, I've made that clear before... we use bucket policies, ACLs are legacy and should never be used.

src/main.js
```diff
@@ -277,13 +279,79 @@ let load = loader({
...
+      };
+
+      if (cfg.app.blobArtifactRegion !== 'us-east-1') {
+        let x = (y) => {
+          y.CreateBucketConfiguration = {
+            LocationConstraint: cfg.app.blobArtifactRegion
+          };
+        }
+        x(publicConfig);
+        x(privateConfig);
```
jonasfj: please, don't do this, I can't read it. Copy/paste is much preferred.

jonasfj: if you really want to do this, use a name that is understandable like `config` instead of `x`..
Or do something like: `[publicConfig, privateConfig].map(config => ....)` but really, I think duplication is the simplest answer here we only have two.

src/main.js
```diff
@@ -277,13 +279,79 @@ let load = loader({
...
+        accessKeyId: cfg.aws.accessKeyId,
+        secretAccessKey: cfg.aws.secretAccessKey,
+      });
+    },
+  },
+
+  // This is a dependency to ease the creation of buckets for, e.g. testing
+  createBuckets: {
+    requires: ['cfg'],
+    setup: async ({cfg}) => {
```
jonasfj: I think this setup method is best split-out into a separate file...

schemas/put-artifact-request.yml
```diff
@@ -0,0 +1,15 @@
+$schema:            http://json-schema.org/draft-04/schema#
+title:              "Put Artifact Request"
```
jonasfj: please name it `Complete Artifact Request` in title and filename. It related to the `completeArtifact` signature...

schemas/put-artifact-response.yml
```diff
@@ -0,0 +1,13 @@
+$schema:            http://json-schema.org/draft-04/schema#
+title:              "Put Artifact Response"
+description: complete a blob artifact upload
+type:               object
+properties:
+  outcome:
+    type: string
+    enum:
+      - success
+      - failure
```
jonasfj: Give a request conflict or some http error if it fails... If you fail to upload then:
A) you gave the wrong etags (400 error)
B) You did upload the parts first (400 error)
C) Something went wrong with the request (500 error)

Client implementors are not likely to handle this error. Note: IMO this doesn't have to provide any response schema, we don't have to provide a response other than 200 OK, or 4xx/5xx...

src/artifacts.js
```diff
@@ -4,8 +4,39 @@ let assert  = require('assert');
...
+  assert(typeof name === 'string');
+  return `${taskId}/${runId}/${name}`;
+}
+
+/**
+ * The reverse of encodeBlobKey.  This will take a key and return an object
+ * with the taskId, runId and name of the artifact.  This method will throw if
+ * the result would not encode into an identical key.
+ */
+function decodeBlobKey(key) {
```
jonasfj: This is never used and has no use-case I can see

src/artifacts.js
```diff
@@ -4,8 +4,39 @@ let assert  = require('assert');
...
 let API     = require('taskcluster-lib-api');
 let urllib  = require('url');
+let crypto  = require('crypto');
 let api     = require('./api');
 
+/**
+ * Encode a storage key for storing in the blob storage type backends We do
+ * this to get better partitioning in S3 which gives us better performance.
+ *
+ * Basically `${md5hash(taskId + runId + name).slice(0,7)}/${taskId}/${runId}/${name}`
```
jonasfj: Bad docs...
Also do we really need this?
The methods is `let encodeBlobKey = (taskId, runId, name) => `${taskId}/${runId}${name}`;
It seems overkill, it's easier to just have the pattern than to type the function name.

src/artifacts.js
```diff
@@ -186,15 +280,20 @@ api.declare({
       break;
 
     case 'azure':
+      present = 1;
```
jonasfj: didn't we move to use booleans?

src/main.js
jonasfj: Btw, when configuring access for the buckets... to the queue, we can use:
http://docs.aws.amazon.com/AmazonS3/latest/API/bucket-policy-s3-sigv4-conditions.html
To create a policy that forbids uploads which don't carry a content signature :)

src/main.js
jonasfj: Also, can't you set the bucket constraint?

jonasfj: Finally, is it perhaps not better to create these buckets and configure them outside the queue code..
There are pros/cons... Amongst the pros is that if we configure it outside the queue code, then the queue needs a lot fewer privileges.. So we can be more strict in the AIM policy we grant the queue.

src/data.js
```diff
@@ -290,6 +290,37 @@ let Artifact = Entity.configure({
...
+    /** 
+     * Present is a number field which represents an artifact being present and
+     * the upload being completed.  The handling logic for the artifact's
+     * storage type will fully define what that means for a given storage type.
+     *
+     * Because there are no boolean fields in Azure-Entities which allow for
+     * fast scanning based on properties (we could use JSON), we're using a
+     * number in place of a boolean.  0 means false, 1 means present
+     */
+    present:        Entity.types.Number,
```
jonasfj: Didn't you say you moved it to use the boolean? or did I miss something... oh, maybe I didn't release that yet..

jonasfj: version 1.4.0 of azure-entities should be out soon, and should include boolean.

schemas/post-artifact-request.yml
```diff
@@ -3,6 +3,110 @@ title:              "Post Artifact Request"
...
+      contentType:
+        description: |
+          Artifact mime-type, when uploading artifact to the signed
+          `PUT` URL returned from this request this must given with the
+           `ContentType` header. Please, provide correct mime-type,
+           this make tooling a lot easier, specifically,
+           always using `application/json` for JSON artifacts.
+        type: string
+        maxLength: 255
+      encoding:
```
jonasfj: Okay, so apart from the naming change... what do we do with `contentEncoding: gzip`, I suspect gzip might be the only content-encoding we want to support? Or do we want `br` support too? Apparently, that's spec too?

Oh, and do we want the sha256 before or after applying `content-encoding`. Reading about `content-md5` it seems that hashing should be done in canonical-form, I understand that as after `content-encoding` have been decoded... Which also makes most sense to me...

But that means we need to send two hashes, one for use as `x-amz-content-sha256` (hash with `content-encoding`) and one for use as `x-amz-meta-content-sha256` (hash without `content-encoding`). For multi-part upload this is no problem, for single part upload we would need a new key, or specify `parts[0] = ...`, and then have the top-level hash be that of the artifact, and the parts hash be those used for transfer.

Hmm, I'm not sure... We might need to pull in some advice from dustin or bstack. I'm not perfectly sure what is best here.

schemas/post-artifact-request.yml
```diff
@@ -3,6 +3,110 @@ title:              "Post Artifact Request"
...
+        description: |
+          The complete SHA256 value of the entire artifact.  This must be the
+          SHA256 of the file which is to be uploaded.  For single part uploads,
+          the upload will fail if the SHA256 value of what is uploaded does not
+          match this value
+        type: string
+        pattern: "[a-fA-F0-9]{64}"
+        minLength: 64
+        maxLength: 64
+      size:
```
jonasfj: Can't we call this `contentLength`

schemas/post-artifact-request.yml
jonasfj: Okay, so I spent a few more thought cycles on this... And talked to bstack.

The natural thing is to have `x-amz-meta-content-sha256` be the hash of the contents in it's canonical form, that is before it is being gzipped.
 * This is what the now obsolute `content-md5` header specifies
 * If making the request through XHR in the browser, the browser will automatically gzip decode, so even if you can access the header you won't be able to verify the hash.

But then we actually do need two different hashes:
 * The canonical hash for `x-amz-meta-content-sha256` (that hash of uncompressed resource)
 * The object hash, or part hash, for `x-amz-content-sha256` to ensure correct transfer.

I propose we do this by changing the data format to be:
```js
{
  contentEncoding: '...', // I propose we limit this to 'gzip', 'br' or 'identity'
  contentType: '...',
  contentSha256: '...', // hash of the entire resource before gzipping
  contentLength: 45345,
  parts: [
    {
      sha256: '...', // 
      size: 345435,
    },
    // One or more parts, but always at-least one.
    // Because we have no use for sha256 of the entire artifact without content-encoding
    // unless it is being uploaded in a single request.
  ],
}
```

jonasfj: I know I'm back to proposing that `parts` is a required property. But it seems we do need two different sha256 hashes, and then it starts to make sense...

jonasfj: Summary:
 * Uploaders can pick gzip or no content-encoding
 * Add `transferSha256` to mean uncompressed hash...

config.yml
```diff
@@ -6,11 +6,16 @@ defaults:
     # Number of seconds before the claim to a run expires
     claimTimeout:                 1200
 
-    # S3 buckets where artifacts are stored
+    # Blob Artifact configuration
+    publicBlobArtifactBucket:     !env PUBLIC_BLOB_ARTIFACT_BUCKET
+    privateBlobArtifactBucket:    !env PRIVATE_BLOB_ARTIFACT_BUCKET
```
jhford: it's in the config.yml file because it's configuration,

config.yml
```diff
@@ -154,6 +162,9 @@ test:
     publishMetaData:              false
     # For testing purposes we let claims expire very fast
     claimTimeout:                 1
+    publicBlobArtifactBucket:     'test-public-blobs'
+    privateBlobArtifactBucket:    'test-private-blobs'
```
jhford: done

test/artifact_test.js
```diff
@@ -70,6 +152,224 @@ suite('Artifacts', function() {
   };
   this.timeout(3 * 60 * 1000);
 
+  suite.only('Blob Storage Type', () => {
+    let tmpobj = tmp.fileSync();
```
jhford: sure.  that doesn't ensure safe generation, but it's also not so big of a deal in this case if it's not safely generated.

jhford: done

package.json
```diff
@@ -53,7 +53,9 @@
     "eslint-plugin-taskcluster": "^1.0.2",
     "mocha": "^2.2.1",
     "mocha-eslint": "^3.0.1",
-    "taskcluster-lib-testing": "^1.0.0"
+    "randbytes": "^0.0.1",
```
jhford: sure!

schemas/artifact-created-message.yml
```diff
@@ -41,8 +41,7 @@ properties:
           artifact.
         type:         string
         enum:
-          - s3
-          - azure
+          - blob
```
jhford: We'd agreed that the message as it was being sent now is actively wrong and has no value and that we should change the message to mean actual completion instead of whatever the current event represents.

Once we've deprecated and blocked new azure/s3, we'll be stopping new artifacts being created with those schemes, so this will hopefully live for a very short amount of time.

schemas/post-artifact-request.yml
```diff
@@ -3,6 +3,110 @@ title:              "Post Artifact Request"
...
+      upload an artifact to storage managed by the queue.
+    type: object
+    properties:
+      storageType:
+        description: Artifact storage type, in this case `'blob'`
+        type: string
+        enum:
+          - blob
+      expires:
+        description: |
```
jhford: I don't have strong feelings on this, but I think that's probably something which we can change later with minimal overhead.  We could start by ignoring the expiration of the artifact message, then eventually remove it from the schema.  I'm worried that we're creeping the scope of this project.

schemas/post-artifact-request.yml
jhford: We had a meeting about this and I sent a follow up email.
```
Hi,

I just wanted to send a quick summary of the meeting we had yesterday
regarding content encoding.  The questions we were trying to address
was which content encodings we should support and whether it would be
mandatory or not.

The consensus reached in the meeting was that we would do the following:

1. support arbitrary content encoding in the code but limit it in
configuration to 'identity' and 'gzip' for now

2. declare that the API consumer must support gzip encoding, but that
the response may or may not be gzip encoded

3. build 'automatic' content-encoding into the uploading and
downloading tools so that the common case of using the corresponding
artifact uploading and downloading tooling can transparently handle
and verify that content-encoding is done correctly

4. we did not come to complete agreement about how the automatic
compressing would decide whether or not to do compression, only that
the decision would be made in the uploading task.

We felt that this approach would cover the majority of use cases.
Simple use cases would use our artifact tooling, and more advanced use
cases could opt to handle the rest request directly.  Since
content-encoding is part of the HTTP spec, it should be reasonably
well supported.

We chose to limit to identity (i.e. no encoding) and gzip encoding
because gzip encoding is prevalent on the internet and well supported
in different programming environments.  We are open to adding other
compressors, but wanted to start with a limited subset at the outset.

We discussed various methods for figuring out whether to compress or
not and felt that the best approach would be to build something
minimally smart into the upload tooling which could be overridden,
while also providing the raw rest api for those who wish to have more
control.

This will not happen until the new artifact api is deployed and the
uploading/downloading tools are provided.

If you have any thoughts or concerns about this, we welcome feedback

Thanks,

John
```

schemas/post-artifact-request.yml
jhford: nice catch, fixed

schemas/post-artifact-request.yml
jhford: Actually, nope.  This is mentioned in the actual JSON Schema documentation that integer in a json schema is not defined and behaves differently depending on the language the validator is implemented on.  This is the suggestion they have to validate in a portable way to deal with integers.

schemas/post-artifact-request.yml
jhford: we can remove it.

src/main.js
jhford: I think it's really nice to have it here because it means that instead of a big script, it's 10 lines of code here.  We're also going to use it so infrequently, I think it's not a huge concern.

src/main.js
jhford: I thought you meant on specific objects.  I'll work on this in a bit since it's a little bigger than a quick change.

src/main.js
jhford: If you'd rather remove it, that's fine with me.

schemas/put-artifact-response.yml
jhford: sometimes clients have bad debugging options and don't expose the status code in an easy way.  the idea here is that we're transferring an extra 20? bytes which could make debugging eaiser.

schemas/post-artifact-request.yml
jhford: i used a short name because it's shorter, i don't really care though.

schemas/post-artifact-request.yml
jonasfj: Spec: http://json-schema.org/latest/json-schema-validation.html#rfc.section.6.25

Note the spec seems to very light on details for most thing.... Regardless we're using integer many places... And it's good for codegen... Granted I guess we can adjust that...

IMO it's also better for docs... And certainly acts as expected...

schemas/put-artifact-response.yml
jonasfj: Because we don't change error codes when you use reply.... Communicate errors with error responses... Please... 

Otherwise tc-client won't throw an exception as expected..

schemas/post-artifact-request.yml
```diff
@@ -3,6 +3,121 @@ title:              "Post Artifact Request"
...
+           always using `application/json` for JSON artifacts.
+        type: string
+        maxLength: 255
+      encoding:
+        description: |
+          Optionally provide an encoding type which should be set as the HTTP
+          Content-Encoding header for this artifact.
+        type: string
+        maxLength: 255
+      sha256:
```
jonasfj: Can we rename this `contentSha256`

schemas/post-artifact-request.yml
```diff
@@ -3,6 +3,121 @@ title:              "Post Artifact Request"
...
+          multipart uploads, this value will be the hash of the overall encoded
+          file that is transfered in portions (e.g. the value of
+          X-Amz-Content-Sha256).  This value is optional when content-encoding
+          is the identity scheme (i.e. no encoding).  This value can be
+          provided with identity encoding if it's valid.  If non-identity
+          encoding is used then the value must be provided
+        type: string
+        pattern: "$[a-fA-F0-9]{64}$"
+        minLength: 64
+        maxLength: 64
```
jonasfj: Regex enforces this... no need for length... same goes for others...

schemas/post-artifact-request.yml
```diff
@@ -3,6 +3,121 @@ title:              "Post Artifact Request"
...
+      size:
```
jonasfj: Can we rename `contentLength`  and `transferLength`

schemas/post-artifact-request.yml
jonasfj: please rename `contentEncoding`

schemas/post-artifact-request.yml
jonasfj: I'm still not convinced we shouldn't make `transferLength` and `transferSha256` required... We can always make things optional later, we never make them required later...

jonasfj:
@jhford, so can we close this PR? ie. it's replaced by the #181, right?

jhford:
yes!

PR #160, opened 2017-04-03
bug 1318648 - support instance instantiation in us-east-2

grenade:
in anticipation of increased load for g2.2xlarge (gecko-t-win7-32-gpu, gecko-t-win10-64-gpu worker types) enable instance instantiation in us-east-2 to mitigate cost and availability issues

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -135,6 +135,7 @@ production:
     taskDependencyTableName:      'QueueTaskDependency'
     cloudMirrorRegions:
       - us-east-1
+      - us-east-2
       - us-west-1
       - eu-central-1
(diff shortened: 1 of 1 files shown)


jonasfj:
@jhford , I think you know best what is needed to support a new region in cloudmirror...

jhford:
So I've made all the changes needed in cloud-mirror, but something is not working correctly.  I've verified that this code works in my development environment, but in production it's not getting copied.  It's like messages from `us-east-2` are getting dropped in production and not making their way to the copier process in the docker-cloud instance.

If we're only planning on using the GPU instances in this region for the time being, we might be able to get by with that minimal load without using cloud-mirror while we debug this.

PR #161, opened 2017-04-03
Fix readme structure section

imbstack:
Just a nice-to-have that I noticed. Please merge if it makes sense to you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,16 +6,11 @@ Project Structure
 -----------------
 _The following itemization of folders outlines how this project is structured._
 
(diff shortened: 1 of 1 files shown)


imbstack:
I think some of the work with artifacts has ended up upgrading records in azure and now this version of the code is trying to run the tests but azure-entities is complaining about versions now. This pollution sucks and our tests should not leak like this. I'll see if I can fix that. To be clear, this is not an issue with the artifacts work, but our tests.

PR #167, opened 2017-05-12
mention the resolve reasons that might trigger a retry, vaguely

djmitche:


schemas/task-status.yml
```diff
@@ -113,7 +113,9 @@ properties:
             Reason that run was resolved, this is mainly
             useful for runs resolved as `exception`.
             Note, **more reasons may be added in the future**, also this
-            property is only available after the run is resolved.
+            property is only available after the run is resolved. Some of these
+            reasons, notably `intermittent-task`, `worker-shutdown`, and
+            `claim-expired`, will trigger an automatic retry of the task.
```
gregarndt: Nice calling that out!

PR #168, opened 2017-05-12
Set clientId for claimed-task temporary creds

djmitche:
This should set the clientId to something like `task-client/HpkPAF3GTkSfA9oxHSHIKQ/0/on/us-west-1/i-09ba01b87daecb599/at/1492905600`.  The `at` part is there so we can debug workers using expired credentials.  I wouldn't mind leaving that off, it just means we'll create a lot of creds with the same clientId.

### DIFF
--- a/src/workclaimer.js
+++ b/src/workclaimer.js
@@ -277,7 +277,18 @@ class WorkClaimer extends events.EventEmitter {
     }, task.routes);
 
     // Create temporary credentials for the task
+    let clientId = [
+      'task-client',
+      taskId,
+      '${runId}',
+      'on',
+      workerGroup,
+      workerId,
+      'until'
+      '${takenUntil.getTime() / 1000}',
+    ].join('/');
     let credentials = taskcluster.createTemporaryCredentials({
+      clientId,
       start:  new Date(Date.now() - 15 * 60 * 1000),
       expiry: new Date(takenUntil.getTime() + 15 * 60 * 1000),
       scopes: [


src/workclaimer.js
```diff
@@ -277,7 +277,18 @@ class WorkClaimer extends events.EventEmitter {
     }, task.routes);
 
     // Create temporary credentials for the task
+    let clientId = [
+      'task-client',
+      taskId,
+      '${runId}',
```
jonasfj: I'm `queue/task/${taskId}/$runId` should be enough...

djmitche:
I suspect we may want to find all authentications by a particular workerId, which is why I included the workerId (and its enclosing workerGroup).

I'm not sure what the choices are that we need to think through?  This really just affects the audit logging, where the more information, the better.

jonasfj:
@imbstack and I tried to sketch out an idea... Where these IDs could be used to trace to grant of authority programmatically... So that one could verify how scopes progress and where an escalation occured...

Granted we only got to the concept stage, and didn't get any further than that :)

Note: workerID, etc can be interred from taskId/runId.

djmitche:
They can be inferred, it's true, but that's not searchable.  I think this clientId could be used to trace the grant, too.  I originally joined the whole thing to `this.credentials.clientId` (which would be `project/taskcluster/tc-queue/production`) but that seemed like a long, useless string so I omitted it.

You're being vague.. have we thought what choice through? What is your idea?  Please be concrete.

jonasfj:
> You're being vague.. have we thought what choice through? What is your idea? Please be concrete.

I have nothing concrete... We only had vague idea :)

IMO it doesn't have to block shipping this.. as we can always change it if we figure out what it is we're rambling about..

djmitche:
@jonasfj is it OK to merge this now, or will that interfere with priority work?

jonasfj:
ship it

Den 16. maj 2017 10.19 AM skrev "Dustin J. Mitchell" <
notifications@github.com>:

djmitche:
Landed 3d06bc3cf0727636185b5ca72fa73e96e3bc6f1a directly -- why doesn't travis build on PRs?

PR #169, opened 2017-05-16
Fix scope issues to require new scopes for priority usage

jonasfj:


src/api.js
```diff
@@ -539,14 +539,15 @@ api.declare({
   stability:  API.stability.stable,
   idempotent: true,
   scopes:     [
+    // Legacy scope options to be removed, hence, no-longer documented
```
imbstack: Let's just delete any unused stuff rather than commenting it out?

jonasfj: yeah, definitely in the future.. I just commented it out for documentation...

PR #175, opened 2017-06-06
Revert "Revert "Limit task.routes to 10""

jonasfj:
Reverts taskcluster/taskcluster-queue#174

Once @escapewindow and @callek are satisfied that patches have been merged around let's land this again...

### DIFF
--- a/schemas/create-task-request.yml
+++ b/schemas/create-task-request.yml
@@ -91,7 +91,7 @@ properties:
       type:         string
       maxLength:    249
       minLength:    1
-    maxItems:       64
+    maxItems:       10 # 4kb frame limit, 10 * 255 = 2550 that leaves a few bytes for other fields
     uniqueItems:    true
   priority:
     title:          "Task Priority"


djmitche:
@jonasfj do we still want to do this?

jonasfj:
No rush to do this soon, I'm okay leaving this hanging.

But we should fix it eventually.

ccooper:
@escapewindow We were waiting on ESR52 to re-land this, but want to make sure that nothing new is using >10 routes. Do you happen to know if that's the case?

escapewindow:
I believe https://dxr.mozilla.org/mozilla-central/source/taskcluster/taskgraph/morph.py#34, along with the functions below it, ensure that no gecko in-tree taskgraph has any task with >10 routes. ESR52 may have either still been using old-style releasetasks, which created the release taskgraph out-of-tree, or may not have had the morph.py changes uplifted.

So I'm fairly confident that gecko taskgraphs are safe. I'm going to guess that no taskcluster-github graph has tasks with over 10 routes, but I don't know how to verify that. It will be easier to say definitively once those are using a standalone taskgraph module that can enforce these things. We're probably ok is my guess.

djmitche:
We're surviving just fine without this change.. let's keep going.

PR #179, opened 2017-07-20
Upgrade to node 8.2

jonasfj:


jonasfj:
Let's ship it as soon as CI passes...

PR #180, opened 2017-07-28
Bug 1384947 - track and list provisioners

djmitche:


schemas/list-provisioners-response.yml
```diff
@@ -0,0 +1,31 @@
...
+properties:
+  provisioners:
+    type:         array
+    items:
+      title:      "Provisioner Information"
+      type:       object
+      properties:
+        provisionerId:
+          title:    "Provisioner ID"
+          type:         string
```
jonasfj:
```
    minLength:      {$const: identifier-min-length}
    maxLength:      {$const: identifier-max-length}
    pattern:        {$const: identifier-pattern}
```

src/workerinfo.js
```diff
@@ -0,0 +1,41 @@
...
+  constructor(options) {
+    assert(options);
+    assert(options.Provisioner);
+    this.Provisioner = options.Provisioner;
+  }
+
+  async provisionerSeen(provisionerId) {
+    let expires = taskcluster.fromNow('5 days');  // temporary hard coded expiration
+
+    let provisioner = await this.Provisioner.load({provisionerId}, true);
```
jonasfj: please store the provisionerId as seen in memory for say 12 hours... that way it won't cost much...
Right now you are adding a request to every single `claimWork` request...

This way the overhead of doing this is `N * |provisionerIds| / 12 hours` where `N` is the number of dynos we have running... But the cost won't grow proportional to the number of requests, which is really important, as we have 20k req/min...

src/api.js
```diff
@@ -1362,9 +1364,12 @@ api.declare({
...
   });
 
-  let result = await this.workClaimer.claim(
-    provisionerId, workerType, workerGroup, workerId, count, aborted,
-  );
+  let [result] = await Promise.all([
+    this.workClaimer.claim(
+      provisionerId, workerType, workerGroup, workerId, count, aborted,
+    ),
+    this.workerInfo.provisionerSeen(provisionerId),
```
jonasfj: Wouldn't it be better to do this as part of `createTask`?

src/workerinfo.js
```diff
@@ -0,0 +1,41 @@
...
+    if (provisioner) {
+      await provisioner.modify(entity => {
+        entity.expires = expires;
+      });
+    } else {
```
jonasfj: Don't make an else-branch, return early and simply the code instead...

src/workerinfo.js
```diff
@@ -0,0 +1,41 @@
...
+    this.Provisioner = options.Provisioner;
+  }
+
+  async provisionerSeen(provisionerId) {
+    let expires = taskcluster.fromNow('5 days');  // temporary hard coded expiration
+
+    let provisioner = await this.Provisioner.load({provisionerId}, true);
+    if (provisioner) {
+      await provisioner.modify(entity => {
+        entity.expires = expires;
```
jonasfj: This is very expensive...
You'll want to do:
```js
await provisioner.modify(provisioner => {
  // Update expiration, if we want to move it by more than 12 hours
  if (new Date(expires) - new Date(provisioner.expires) > 12 * 60 * 60 * 1000) {
    provisioner.expires = expires;
  }
})
```

Now it won't have to update more than 12 hours, sure expiration might be off by like, 12 hours,  but given that we want it to be something like 5 days anyways that doesn't really matter...

djmitche:
> This is not feasible, too many requests...

Well, it's feasible, just not with this approach.  Thanks for spotting that!

src/api.js
djmitche: I think we want to monitor the provisioners and workers that are out there ready to do work, rather than the provisioners and workers for which someone has called createTask.

schemas/list-provisioners-response.yml
djmitche: fixed

src/api.js
jonasfj: We could also choose to not block on `provisionerSeen`... it call it but not `await` it... we would obviously have to report errors directly to sentry then...

src/api.js
```diff
@@ -1969,6 +1977,45 @@ require('./artifacts');
...
+    'option. By default this end-point will list up to 1000 provisioners in a single',
+    'page. You may limit this with the query-string parameter `limit`.',
+  ].join('\n'),
+}, async function(req, res) {
+  let continuation = req.query.continuationToken || null;
+  let limit = parseInt(req.query.limit || 1000, 10);
+
+  let provisioners = await this.Provisioner.scan({}, {continuation, limit});
+  let result = {
+    provisioners: provisioners.entries.map(artifact => artifact.json()),
```
jonasfj: the name `artifact` is that a copy/paste thing :)

src/data.js
```diff
@@ -613,3 +613,46 @@ TaskDependency.expire = async function(now) {
...
+    limit:            250, // max number of concurrent delete operations
+    handler:          entry => { count++; return entry.remove(true); },
+  });
+  return count;
+};
+
+/** Return JSON representation of provisioner meta-data */
+Provisioner.prototype.json = function() {
+  return {
+    provisionerId:    this.provisionerId,
```
jonasfj: I **love** this indentation!! :)

This is so awesome.. everybody said I was crazy for aligning everything... I'm so happy to see that I'm not alone anymore! :)

src/workerinfo.js
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * Mark a value as seen, only actually updating the Azure table row
+   * occasionally. This takes the approach of updating the row immediately on
+   * the first call, then not updating until `updateFrequency` has elapsed.
+   * Thus the `expires` value on a row may be out-of-date by `updateFrequency`.
+   *
+   * Note that the cache is never purged of outdated entries; this assumes that
+   * the process is restarted on a daily basis, so there is not too much time
+   * for stale cache entries to accumulate.
```
jonasfj: How about we keep a `this._purgeCacheAt` time, and then we remove expired entries every 60min or so... by checking  `this._purgeCacheAt` every time `valueSeen` is called...

src/workerinfo.js
```diff
@@ -0,0 +1,69 @@
...
+   */
+  async valueSeen(key, updateExpires) {
+    let now = new Date();
+    let nextUpdate = this.nextUpdateAt[key];
+    if (!nextUpdate || nextUpdate < now) {
+      await updateExpires();
+      this.nextUpdateAt[key] = taskcluster.fromNow(this.updateFrequency);
```
jonasfj: Do this before `await updateExpires();` to avoid doing it multiple times in parallel...

Notably, wrap `await updateExpires();` with a try/catch, where you reset `this.nextUpdateAt[key]` if an exception is thrown... well, maybe, if we can't update it maybe it's better to send the error to sentry and not cause 500 errors on the queue...

src/workerinfo.js
```diff
@@ -0,0 +1,69 @@
...
+
+  async provisionerSeen(provisionerId) {
+    await this.valueSeen(provisionerId, async () => {
+      let expires = taskcluster.fromNow('5 days');  // temporary hard coded expiration
+
+      // perform an Azure upsert, trying the update first as it is more common
+      let provisioner = await this.Provisioner.load({provisionerId}, true);
+      if (provisioner) {
+        await provisioner.modify(entity => {
+          entity.expires = expires;
```
jonasfj: Again you'll want do a conditional modification here... no need  that every dyno has to update every entry very 6 hours...
Notably this would cause some racing...

To avoid all that do:
```js
if (Date.now() - new Date(entity.expires) > 24 * 60 * 60 * 1000) {
  entity.expires = taskcluster.fromNow('5 days');
}
```

That way we'll only update once every day.

src/data.js
djmitche: No, you're still alone, but when in Denmark..

src/api.js
helfi92: How do you report errors directly to sentry?

helfi92:
Continued in #182

src/api.js
jhford: you would call `monitor.reportError`

Not having the promise `await`ed or `.catch()`'d seems like a bad idea, especially as it completely drops on the floor if something goes wrong.  What about storing a reference to the promise and waiting for it after the res.reply?

djmitche: We talked a little about this in irc and decided to leave it as-is for now: running the claim operation and the provisionerSeen operation in parallel.  In most cases the provisionerSeen will just be checking a timestamp, so it will be much faster, and anyway claim is generally long-polling.

PR #181, opened 2017-08-03
Blob artifacts q3

jhford:
Patch rebased.  I still need to do the four renames we discussed.


jhford:
I've addressed the feedback we agreed on.

jhford:
@jonasfj let me know if there's anything else.

src/main.js
```diff
@@ -308,13 +312,79 @@ let load = loader({
     },
   },
 
+  s3Controller: {
+    requires: ['cfg'],
+    setup: async ({cfg}) => {
+      return new remoteS3.Controller({
```
jhford: we don't want to create buckets every time we launch the app

src/data.js
```diff
@@ -348,6 +348,38 @@ let Artifact = Entity.configure({
...
   ],
+}).configure({
+  version: 2,
+  properties: {
+    taskId:         Entity.types.SlugId,
+    runId:          Entity.types.Number,
+    name:           Entity.types.String,
+    storageType:    Entity.types.String,
+    contentType:    Entity.types.String,
+    details:        Entity.types.JSON,
```
jhford: sure.  fwiw, that's explained already in `artifact.js`

src/data.js
```diff
@@ -348,6 +348,38 @@ let Artifact = Entity.configure({
...
+    expires:        Entity.types.Date,
+    /** 
+     * Present is a number field which represents an artifact being present and
+     * the upload being completed.  The handling logic for the artifact's
+     * storage type will fully define what that means for a given storage type.
+     *
+     * Because there are no boolean fields in Azure-Entities which allow for
```
jhford: yes, and I use it four lines down.

schemas/post-artifact-request.yml
```diff
@@ -3,6 +3,116 @@ title:              "Post Artifact Request"
...
+              pattern: "^[a-fA-F0-9]{64}$"
+              minLength: 64
+              maxLength: 64
+            size:
+              description: |
+                The number of bytes in this part.  Keep in mind for S3 that
+                all but the last part must be minimum 5MB and the maximum for
+                a single part is 5GB.  The overall size may not exceed 5TB
+              type: number
+              multipleOf: 1.0
```
jhford: but the docs specifically say that `type: integer` does not ensure cross-platform/language behaviour and specifically says to use this to ensure an actual integer

src/data.js
```diff
@@ -348,6 +348,63 @@ let Artifact = Entity.configure({
...
+     * storageType: error
+     *   reason:        Formalized reason for error artifact, see JSON schema.
+     *   message:       Human readable error message to return
+     * storageType: blob
+     *   contentType
+     *   contentEncoding
+     *   contentSha256: SHA256 hash of the un-encoded artifact
+     *   contentLength: Number of bytes of the un-encoded artifact
+     *   transferSha256: SHA256 hash of the content-encoding encoded artifact
+     *   transferLength: Number of bytes of the content-encoding encoded artifact
```
jonasfj: So `transferSha256` and `transferLength` are optional? this is something we should be sure to document..

schemas/post-artifact-request.yml
jonasfj: What docs? The spec says no such thing: http://json-schema.org/latest/json-schema-validation.html#rfc.section.6.25

Note: We use `type: integer` extensively...

The only cross platform issues I can imagine is whether the representation is  float64, int32 or int64, but that's exclusively and implementation issue.. And as long as we're below `2^54` which is the biggest int node can handle anyways, I think we'll be fine..

Note: JSON spec doesn't say what a number is either: http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf
And arguably whether you represent JSON in memory as float64 or int64 is up to the implementor... 

Please reference sources, we're using `type: integer` a lot, everywhere and it makes codegen much nicer...

schemas/post-artifact-response.yml
```diff
@@ -5,6 +5,52 @@ description: |
...
+          Artifact storage type, in this case `'blob'`
+        type: string
+        enum:
+          - blob
+      expires:
+        description: |
+          Date-time after which the signed `requests` no longer work
+        type: string
+        format: date-time
+      requests:
```
jonasfj: Couldn't we rename this `parts` ? and give each part a properties `offset` and `size`?

jhford:
@jonasfj I think I've addressed the feedback so far.  When do you think you'll have a chance to go through the `artifact.js` logic again?

src/data.js
```diff
@@ -348,6 +348,63 @@ let Artifact = Entity.configure({
...
+     *   partsHash: Rather than storing the potentially large list of sha256/size
+     *              pairings for each part, we just need to store enough information
+     *              to determine if the operation would result in an idempotency
+     *              error
+     */
```
jonasfj: Please document the `uploadId` property... When it's present, and what it is...

jonasfj: Please document the `region` and `provider` properties

jonasfj: Please document the `bucket` and `key` properties... anything else we're missing? Please check it's super hard to maintain this if we forget what properties we're storing.

src/data.js
jonasfj: Or maybe we should make `transferSha256 == contentSha256` if there is no difference... ie. let it default to `contentSha256` if not given... same for `transferLength`?

src/artifacts.js
```diff
@@ -423,6 +643,99 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
 
+/** Complete artifact */
+api.declare({
+  method:     'put',
+  route:      '/task/:taskId/runs/:runId/artifacts/:name(*)',
+  name:       'completeArtifact',
+  stability:  API.stability.experimental,
+  scopes: [
+    [
+      'queue:create-artifact:<name>',
```
jonasfj: This is a new end-point no need to keep legacy scopes... please remove this option...

src/artifacts.js
```diff
@@ -423,6 +643,99 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+      artifact: artifact.json(),
+      workerGroup,
+      workerId,
+      runId,
+    }, task.routes);
+    return res.reply({
+      outcome: 'success',
+    });
+  } else {
+    throw new Error('Only supported storageType artifacts may be marked completed');
```
jonasfj: Don't make an else-branch, prefer to return early...Also this causes a 500-error which is wrong.
This should be a 400 error! See: https://github.com/taskcluster/taskcluster-lib-api/blob/master/src/errors.js#L9
`res.reportError('InputError', 'queue.completeArtifact should only be called for "blob" artifacts', {artifact: artifact.json()})`

src/artifacts.js
```diff
@@ -423,6 +643,99 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+    this.Artifact.load({taskId, runId, name}, true),
+    this.Task.load({taskId}, true),
+  ]);
+
+  // TODO: Have a shortcut here where things which are already
+  // marked as present short circuit
+
+  let run = task.runs[runId];
+  if (!run) {
+    return res.reportError('InputError',
```
jonasfj: This is a `ResourceNotFound` error...

src/artifacts.js
```diff
@@ -423,6 +643,99 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+  let name        = req.params.name;
+  let input       = req.body;
+
+  // Hmm, list destructuring is probably better here
+  let [artifact, task] = await Promise.all([
+    this.Artifact.load({taskId, runId, name}, true),
+    this.Task.load({taskId}, true),
+  ]);
+
+  // TODO: Have a shortcut here where things which are already
```
jonasfj: Yeah, if `artifact.present == true` then return early...

src/artifacts.js
```diff
@@ -423,6 +643,99 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+      'assume:worker-id:<workerGroup>/<workerId>',
+    ], [
+      'queue:create-artifact:<taskId>/<runId>',
+    ],
+  ],
+  deferAuth:  true,
+  input:      'put-artifact-request.json#',
+  output:     'put-artifact-response.json#',
+  title:      'Complete Artifact',
+  description: 'tbd',
```
jonasfj: Add description... outline when this should be called, it's only for `blob` artifacts right?

src/artifacts.js
```diff
@@ -423,6 +643,99 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+
+    await this.publisher.artifactCreated({
+      status: task.status(),
+      artifact: artifact.json(),
+      workerGroup,
+      workerId,
+      runId,
+    }, task.routes);
+    return res.reply({
+      outcome: 'success',
```
jonasfj: Just return `res.reply({})` or maybe just `res.status(200).send()`, don't send a dummy value... who knows in some future we might want to actually return something useful...

src/artifacts.js
```diff
@@ -423,6 +643,99 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+      etag = input.etags[0];
+    }
+    // TODO: Do a quick HEAD of the resource (maybe) to double check that the
+    // expected x-amz-meta-content-sha256 value matches and that the etag we
+    // either received or generated by completing a multipart upload matches
+    // the expected value
+    await artifact.modify((artifact) => {
+      artifact.details.etag = etag;
+      // Now that we're finished, we don't want to store the uploadId any longer
+      artifact.details = _.omit(artifact.details, 'uploadId');
```
jonasfj: This invariant should be documented in `data.js`

src/artifacts.js
```diff
@@ -423,6 +643,99 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+      });
+    } else {
+      etag = input.etags[0];
+    }
+    // TODO: Do a quick HEAD of the resource (maybe) to double check that the
+    // expected x-amz-meta-content-sha256 value matches and that the etag we
+    // either received or generated by completing a multipart upload matches
+    // the expected value
+    await artifact.modify((artifact) => {
+      artifact.details.etag = etag;
```
jonasfj: `etag` property is not documented in `data.js`, please ensure that it's documented and that it's documented when it's supposed to be present...

src/artifacts.js
```diff
@@ -423,6 +643,99 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+    if (artifact.details.partsHash) {
+      etag = await this.s3Controller.completeMultipartUpload({
+        bucket: artifact.details.bucket,
+        key: artifact.details.key,
+        etags: input.etags,
+        uploadId: artifact.details.uploadId,
+        tags: {taskId, runId, name},
+      });
+    } else {
+      etag = input.etags[0];
```
jonasfj: This is where we're supposed to do a `HEAD` request, otherwise the `artifactCreated` message has no value...

In fact we don't promise that artifacts are present if we don't this here....

src/artifacts.js
```diff
@@ -423,6 +643,99 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+    }
+    // TODO: Do a quick HEAD of the resource (maybe) to double check that the
```
jonasfj: The `HEAD` request should be done in else-branch above...

src/data.js
jonasfj: Document the `etag` property

jonasfj:
I hate to expand the scope of this PR... but I think we're missing something. We had 3 goals:
 * Multi-part upload
 * Content hashing and verification
 * Ensure that artifacts are successfully upload if present

So in thinking about I think we should modify `api.js` such that `resolveTask` refuses the resolve a run as successful if blob artifact uploads have been started but not completed...

This isn't actually very hard to do in:
https://github.com/taskcluster/taskcluster-queue/blob/eda08ebdd707498fb64658ac5e6c158719bd808a/src/api.js#L1679-L1681

we just do a query to see if there is any artifacts with `{storageType: 'blob', present: false}` like this:
```js
var resolveTask = async function(req, res, taskId, runId, target) {
   ...
   let task = await this.Task.load({taskId}, true);
   ...

  if (target == 'completed') {
    let artifactsCompleted = true;
    await this.artifacts.query({taskId, runId, storageType: 'blob', present: false}, {
      limit: 1,
      handler: () => {artifactsCompleted = false;},
    });
    if (!artifactsCompleted) {
      // TODO: Resolve task exception 'internal-error' or add a new exception reason 'incomplete-artifacts'
      return res.reportError('InputError', 'artifacts must be completed prior to reportCompleted');
    }
  }

  ...
};
```

This way we can guarantee that tasks that are reported completed will have their artifacts. I think that's super useful invariant that we can rely on many places.

jonasfj:
@jhford, hi, I think you forgot to push last time you asked me for more feedback..
Anyways, take a look I found something we should document...

I suspect that in the long run it'll serve us well if the contents of the `details` property is documented in a single place. Otherwise, we might end up having to do some guess work about what is in our data store, that wouldn't be fun.

PR #183, opened 2017-08-04
Bug 1384948 - API endpoint to list out worker types for a given provisioner

helfi92:


helfi92:
@djmitche Would it be possible to look at the commit so far and give some feedback. Also, do I need to handle expiration? If I do, how should I go into handling expiry in worker-type entries

src/api.js
```diff
@@ -1369,6 +1370,7 @@ api.declare({
       provisionerId, workerType, workerGroup, workerId, count, aborted,
     ),
     this.workerInfo.provisionerSeen(provisionerId),
+    this.WorkerType.create({provisionerId, workerType}, true),
```
djmitche: I think this needs to use the same "seen" logic as provisioners - so something like `this.workerInfo.workerTypeSeen`.

src/api.js
```diff
@@ -2046,3 +2048,50 @@ api.declare({
...
+  query: {
+    continuationToken: /./,
+    limit: /^[0-9]+$/,
+  },
+  name:       'listWorkerTypes',
+  stability:  API.stability.experimental,
+  output:     'list-workertypes-response.json#',
+  title:      'Get a list of all active worker-types',
+  description: [
+    'Get all active worker-types.',
```
djmitche: ..for the given provisioner.

src/api.js
```diff
@@ -2046,3 +2048,50 @@ api.declare({
...
+  const provisionerId = req.params.provisionerId;
+  let workerTypes = await this.WorkerType.scan({provisionerId}, {continuation, limit});
+
+  let result = {
+    workerTypes: workerTypes.entries.map(workerType => workerType.workerType),
+  };
+
+  // If no taskGroup was found
+  if (!result.workerTypes.length) {
+    return res.reportError('ResourceNotFound',
```
djmitche: I think returning an empty array here would be OK.  At least, that would be accurate for provisioners that currently have no active workerTypes, even if it's inaccurate when `provisionerId` doesn't exist.  I suppose the most correct choice would be to consult the Provisioner table in this case, and return 404 if there's no such provisioner, otherwise [].  But I'm comfortable with just returning [] unconditionally, too.

src/data.js
```diff
@@ -656,3 +656,19 @@ Provisioner.prototype.json = function() {
 
 // Export Provisioner
 exports.Provisioner = Provisioner;
+
+/**
+ * Entity for tracking worker-types.
+ */
+let WorkerType = Entity.configure({
+  version:            1,
+  partitionKey:       Entity.keys.StringKey('workerType'),
```
djmitche: Per https://bugzilla.mozilla.org/show_bug.cgi?id=1384947#c1 I think that should be PartitionKey=provisionerId,            RowKey=workerType (both StringKey, meaning get that value from the properties and use it as the key).

test/runtests.sh
```diff
@@ -27,4 +26,6 @@ mocha                             \
   .test/expiretask_test.js        \
   .test/expirequeues_test.js      \
   .test/dependency_test.js        \
+  .test/workerinfo_test.js        \
```
djmitche: Oh, good catch -- most services have been updated to just use `.test/*_test.js` but I guess this one got missed..

helfi92:
@djmitche Do I need to handle some type of expiration? if so, how would you go into handling expiry in worker-type entries

djmitche:
Yes, expiration should be handled in the same way that provisioner expiration is handled -- add some code to the `WorkerInfo.expire` method to call `WorkerType.expire`.

helfi92:
Sounds good, thanks!

test/workerinfo_test.js
```diff
@@ -1,49 +1,120 @@
...
+      continuationToken: result.continuationToken,
+    });
+
+    assert(!result.continuationToken);
+    assert(result.workerTypes.length === 1);
+  });
+
+  test('workerTypeSeen creates and updates a worker-type', async () => {
+    let workerInfo = await helper.load('workerInfo', helper.loadOptions);
+    let workerTypeA = slugid.v4();
```
djmitche: it doesn't really matter, since this is a test, but workerTypes aren't slugids -- they're things like "github-worker" or "gecko-b-2-linux".

helfi92: Right, thanks for the reminder :) I used slugids since this is how I saw it being tested in other test files.

djmitche: Probably . I think that's a leftover from when tests didn't clean up after themselves.

helfi92: Ah I see. In that case, I will fix this test file to have better worker-type names.

helfi92:
@djmitche Safe to merge? :)

djmitche:
Well, I thought *my* patch was safe to merge, and it wasn't, so I don't know if I have any authority to say so!  But yes, I think so.  It might be wise to wait a few hours until @jonasfj is around in case something goes sideways.

config.yml
```diff
@@ -21,6 +21,7 @@ defaults:
     taskGroupActiveSetTableName:  TaskGroupActiveSets
     taskRequirementTableName:     TaskRequirement
     taskDependencyTableName:      TaskDependency
+    workerTypeTableName:          WorkerType
```
jonasfj: plural?

config.yml
```diff
@@ -21,6 +21,7 @@ defaults:
...
     provisionerTableName:         Provisioner
```
jonasfj: I guess we forgot that here... we can change it... no harm in leaving the old table there... or we can just keep it, no worries :)

schemas/list-workertypes-response.yml
```diff
@@ -0,0 +1,31 @@
...
+    type:           array
+    title:          "WorkerTypes from the Provisioner"
+    description: |
+      List of worker-types in this provisioner.
+    items:
+      type:         string
+      title:        "WorkerType name"
+      description: |
+        WorkerType name.
+      additionalProperties: false
```
jonasfj: See task definition for regex pattern for these identifiers...

jonasfj: Also maybe instead of making it a list of strings... let's make it a list of objects... that way we can add additional properties for each workerType in the future (without breaking compatibility).

Ie:
```js
{
 workerTypes: [
    {workerType: '...'},
    ....
  ]
}
```

src/api.js
```diff
@@ -2046,3 +2048,41 @@ api.declare({
...
+    'Get all active worker-types for the given provisioner.',
+    '',
+    'The response is paged. If this end-point returns a `continuationToken`, you',
+    'should call the end-point again with the `continuationToken` as a query-string',
+    'option. By default this end-point will list up to 1000 worker-types in a single',
+    'page. You may limit this with the query-string parameter `limit`.',
+  ].join('\n'),
+}, async function(req, res) {
+  let continuation = req.query.continuationToken || null;
+  let limit = parseInt(req.query.limit || 1000, 10);
```
jonasfj: max 1000, maybe that needs to be fixed for listProvisioners too?

src/workerinfo.js
```diff
@@ -59,12 +61,43 @@ class WorkerInfo {
     });
   }
 
+  async workerTypeSeen(provisionerId, workerType) {
+    await this.valueSeen(workerType, async () => {
```
jonasfj: `workerType` is not globally unique... you can do: `provisionerId + '/' + workerType` as the key... since neither can contain `/`

src/workerinfo.js
```diff
@@ -59,12 +61,43 @@ class WorkerInfo {
     });
   }
 
+  async workerTypeSeen(provisionerId, workerType) {
```
jonasfj: Maybe `WorkerInfo` just has `WorkerInfo.seen(provisionerId, workerType)` which then in turn calls the two different seen functions concurrently...

schemas/list-workertypes-response.yml
djmitche: ++ I did this for the provisioners too.

config.yml
djmitche: Most of them are singular -- that's the pattern I followed with "Provisioner"

jonasfj: I guess we could just choose not to care, rather than debate this :)

I think `TaskRequirement` and `TaskDependency` are singular because they are relations, more than they are data-tables...

src/api.js
```diff
@@ -2001,11 +2000,12 @@ api.declare({
     'page. You may limit this with the query-string parameter `limit`.',
   ].join('\n'),
 }, async function(req, res) {
-  let continuation = req.query.continuationToken || null;
-  let limit = parseInt(req.query.limit || 1000, 10);
+  const continuation = req.query.continuationToken || null;
+  const isLimitValid = req.query.limit && req.query.limit < 1000;
+  const limit = parseInt(isLimitValid ? req.query.limit : 1000, 10);
```
jonasfj: Just do:
`const limit = Math.min(1000, parseInt(req.query.limit || 1000, 10));`

Notice that: `req.query.limit < 1000` is actually string and int comparison...It works, but let's avoid it :)

helfi92: hehe, nice stamp. Yeah I like your way much better. Will fix :)

PR #184, opened 2017-08-09
Bug 1384949 - Implement API endpoint to list out a list of workerGroup/workerId

helfi92:


schemas/list-workers-response.yml
```diff
@@ -0,0 +1,43 @@
...
+    type:           array
+    title:          "Workers from a WorkerType"
+    description: |
+      List of workers in this worker-type.
+    items:
+      type:         object
+      properties:
+          workerGroup:
+                    description: |
+                      Identifier for the worker-group within which this run started.
```
djmitche: This isn't really related to a "run".  How about "Identifier for the worker group containing this worker"

schemas/list-workers-response.yml
```diff
@@ -0,0 +1,43 @@
...
+                    type:         string
+                    minLength:    {$const: identifier-min-length}
+                    maxLength:    {$const: identifier-max-length}
+                    pattern:      {$const: identifier-pattern}
+          workerId:
+                    description: |
+                      Identifier for the worker executing this run.
```
djmitche: "Identifier for this worker (unique within this worker group)"

src/workerinfo.js
```diff
@@ -34,57 +35,65 @@ class WorkerInfo {
...
+      try {
+        await entity.create(entry);
+      } catch (err) {
+        // EntityAlreadyExists means we raced with another create, so just let it win
+        if (!err || err.code !== 'EntityAlreadyExists') {
+          throw err;
+        }
+      }
+    };
+
```
djmitche: nice refactor :)

src/workerinfo.js
```diff
@@ -34,57 +35,65 @@ class WorkerInfo {
     }
   }
 
-  async seen(provisionerId, workerType) {
+  async seen(provisionerId, workerType, workerGroup, workerId) {
     const expired = entity => Date.now() - new Date(entity.expires) > 24 * 60 * 60 * 1000;
     const expires = taskcluster.fromNow('5 days');  // temporary hard coded expiration
```
djmitche: I think we need to have a much shorter expires for workers than for workerTypes or provisioners, since we create so many transient workers in EC2.  @jonasfj do you have a good expiration in mind? This could be very short if we also call workerInfo.seen on reclaimTask..

jonasfj:
I'm wondering if maybe this would be better in postgres where we can query it?

djmitche:
@jonasfj yes maybe, but we don't want to block this project on using postgres.

src/workerinfo.js
djmitche: Now that I think about this some more, there are a few things that will prevent that:
* expiration only runs daily
* expiration is only updated in the table every 6 hours

So I think the minimum expiration is one day.

In fact, looking at
```js
 const expired = entity => Date.now() - new Date(entity.expires) > 24 * 60 * 60 * 1000;
```
I think we wait until 24 hours *after* an entry is expired to write its new expiration date?  Shouldn't that be the reverse?

src/workerinfo.js
```diff
@@ -34,57 +35,65 @@ class WorkerInfo {
...
 
+    const updateExpiration = (entry) => {
+      entry.modify(entity => {
+        if (expired(entity)) {
+          entity.expires = expires;
+        }
+      });
+    };
+
+    const createEntry = async (entity, entry) => {
```
jonasfj: this is nit... but I would suggest a global utility like:
```js
let ignoreEntityAlreadyExists = (err) => {
    throw err;
  }
}
```

Then use it like:
```js
await this.workerType.create({provisionerId, workerType, expires}).catch(ignoreEntityAlreadyExists);
```

src/workerinfo.js
```diff
@@ -34,57 +35,65 @@ class WorkerInfo {
     }
   }
 
-  async seen(provisionerId, workerType) {
+  async seen(provisionerId, workerType, workerGroup, workerId) {
     const expired = entity => Date.now() - new Date(entity.expires) > 24 * 60 * 60 * 1000;
     const expires = taskcluster.fromNow('5 days');  // temporary hard coded expiration
 
+    const updateExpiration = (entry) => {
```
jonasfj: We could also do this as a global helper... like:
```js
await workerType.modify(updateExpires);
```
or something like: 
```js
await workerType.modify(updateExpires(offset, ...));
```

On both cases `updateExpires` would be a global helper function defined inside the file and not exported...

src/workerinfo.js
helfi92: @djmitche Agreed, we are currently waiting 24 hours after an entry is expired to write the updated value. I am assuming we want to update the expiration value if there's less than 24 hours left to expiration. Also, is 3 days for the expiration time short enough for workers?

```js
const expires = workerId ? taskcluster.fromNow('3 days') : taskcluster.fromNow('5 days');
const expired = entity => new Date(entity.expires) - Date.now() < 24 * 60 * 60 * 1000;
```

src/workerinfo.js
helfi92: @jonasfj From my understanding, the `await` operator returns the resolved value of a promise so chaining the catch method will probably not work. I think we'll have to use try and catch statements:
```js
const createEntry = async (entity, entry) => {
  try {
    await entity.create(entry);
  } catch (err) {
    ignoreEntityAlreadyExists(err);
  }
};
```

src/workerinfo.js
```diff
@@ -34,57 +35,65 @@ class WorkerInfo {
...
-          // EntityAlreadyExists means we raced with another create, so just let it win
-          if (!err || err.code !== 'EntityAlreadyExists') {
-            throw err;
-          }
+        createEntry(this.WorkerType, {provisionerId, workerType, expires});
+      });
+    };
+
+    const workerSeen = async (provisionerId, workerType, workerGroup, workerId) => {
+      await this.valueSeen(`${workerGroup}/${workerId}`, async () => {
```
helfi92: @jonasfj workerTypeSeen is already nesting under `${provisionerId}/${workerType}`. Wouldn't that interfere with the proposed nesting? I am thinking we could do the nesting under `${provisionerId}/${workerType}/${workerGroup}/${workerId}`. What do you think?

src/workerinfo.js
```diff
@@ -95,6 +104,10 @@ class WorkerInfo {
     if (provisionerId && workerType) {
       workerTypeSeen(provisionerId, workerType);
     }
+
+    if (provisionerId && workerType && workerGroup && workerId) {
+      workerSeen(provisionerId, workerType, workerGroup, workerId);
```
helfi92: Good catch!

src/api.js
```diff
@@ -2084,3 +2085,46 @@ api.declare({
...
+    'option. By default this end-point will list up to 1000 workers in a single',
+    'page. You may limit this with the query-string parameter `limit`.',
+  ].join('\n'),
+}, async function(req, res) {
+  const continuation = req.query.continuationToken || null;
+  const provisionerId = req.params.provisionerId;
+  const workerType = req.params.workerType;
+  const limit = Math.min(1000, parseInt(req.query.limit || 1000, 10));
+
+  const workers = await this.Worker.scan({provisionerId, workerType}, {continuation, limit});
```
helfi92: Agreed 👌

src/workerinfo.js
djmitche: Let's do a 1-day expiration for workers.  And yes, let's fix that comparison :)

src/workerinfo.js
djmitche: I think that's exactly what he meant.  It's what I would expect anyway :)

src/workerinfo.js
djmitche: Much shorter than 1 day risks a worker expiring while running a long task.  Ideally at some point we'll call workers "seen" on reclaimWork, at which point we could reduce this to an hour or so, but not yet.

djmitche:
I'm a little worried about performance, too.  @jonasfj you have a load-testing framework set up for the queue, right?  Is there a way to test this with that framework?

helfi92:
I ran `heroku run --app tc-queue-load-test -- node lib/main load-test` on 3 terminals in parallel on a single dyno. The tc-queue-load-test [logs](https://papertrailapp.com/systems/tc-queue-load-test/events) on papertrail seems to be ok.

djmitche:
:shipit:

PR #185, opened 2017-08-10
add expireWorkerInfo to Procfile

djmitche:


djmitche:
I just added this to Heroku, but it will fail until this Procfile update

PR #186, opened 2017-08-16
Bug 1390865 - Expand the /provisioners endpoint to return metadata

helfi92:


src/data.js
```diff
@@ -627,6 +627,8 @@ let Provisioner = Entity.configure({
     provisionerId:    Entity.types.String,
     // the time at which this provisioner should no longer be displayed
     expires:          Entity.types.Date,
+    description:      Entity.types.String,
+    stability:        Entity.types.String,
```
jonasfj: You can't add new fields... you must make a new version.

schemas/list-provisioners-response.yml
```diff
@@ -9,14 +9,36 @@ properties:
...
+          title:        "Provisioner ID"
           type:         string
           minLength:    {$const: identifier-min-length}
           maxLength:    {$const: identifier-max-length}
           pattern:      {$const: identifier-pattern}
+        stability:
+          title:        "Stability"
+          description: |
+            This is the stability of the provisioner. Accepted values:
+            'experimental|stable|deprecated'
```
jonasfj: descriptions in our stuff can be markdown, so:
```markdown
This is the stability of the provisioner. Accepted values:
 * `experimental`
 * `stable`
 * `deprecated`
```

that way it'll look nicer on the docs site...

PR #187, opened 2017-08-17
Bug 1390867 - Expand the worker-types endpoint to return metadata

helfi92:


src/workerinfo.js
```diff
@@ -87,7 +87,15 @@ class WorkerInfo {
...
 
-        createEntry(this.WorkerType, {provisionerId, workerType, expires});
+        createEntry(this.WorkerType, {
+          provisionerId,
+          workerType,
+          expires,
+          description: '',
+          stability: 'experimental',
+          documentation: '',
+          payloadSchema: {},
```
djmitche: this should probably be `null` to indicate no schema is defined - I think `{}` only matches an empty object.

schemas/list-workertypes-response.yml
```diff
@@ -20,10 +20,45 @@ properties:
...
+            The stability of the worker-type. Accepted values:
+            'experimental|stable|deprecated'
+          type:         string
+          enum:         ["experimental", "stable", "deprecated"]
+        description:
+          title:        "Description"
+          description: |
+            Description of the worker-type.
+          type:         string
+        documentation:
```
djmitche: This is the list resource.  Would it make sense to omit documentation and payload schema here, but include them a get-workertype endpoint?  So, same REST resource, just different views?

helfi92: @djmitche What do you mean with "same REST resource, just different views". Did you mean the following:
`GET /v1/listWorkerTypes` omits documentation and payload
`GET /v1/provisioners/:provisionerId/worker-types/:workerType` includes documentation and payload

If that's not what you meant, would ^ be a viable option? In that case, `GET /v1/provisioners/:provisionerId/worker-types/` would perhaps also omit documentation and payload.

djmitche: Yes, that's what i meant, although I think the paths are actually
```
GET /v1/provisioners/:provisionerId/worker-types  # list of workerTypes without details
GET /v1/provisioners/:provisionerId/worker-types/:workerType  # single workerType with details
```

so from a REST perspective, those are the same thing, just one view has more detail than the other.

helfi92: I think that's a great idea!

helfi92:
Since we agreed that we will store the derived claimWork data in azure-entity and everything else on azure-blob or s3-object, I will go ahead and remove `description`, `stability`, `documentation` and `payloadSchema`from the azure-entities.

helfi92:
Pushed a commit that stores the worker-type metadata on azure-blob. I'm still unsure if this is the route we want to take so no need to review for now...

helfi92:
I rebased on my old commit which had the metadata stored in the same entity.

src/data.js
```diff
@@ -672,6 +672,26 @@ let WorkerType = Entity.configure({
...
     expires:          Entity.types.Date,
   },
+}).configure({
+  version:            2,
+  properties: {
+    provisionerId:    Entity.types.String,
+    workerType:       Entity.types.String,
+    // the time at which this worker-type should no longer be displayed
+    expires:          Entity.types.Date,
+    description:      Entity.types.String,
```
jonasfj: Type text might be better... As it can contain more data

src/data.js
```diff
@@ -672,6 +672,26 @@ let WorkerType = Entity.configure({
...
+    stability:        Entity.types.String,
+    documentation:    Entity.types.String,
+    payloadSchema:    Entity.types.JSON,
```
jonasfj: Docs and schema should not be in azure-entities, S3 or blob storage is probably best

src/data.js
```diff
@@ -672,6 +672,26 @@ let WorkerType = Entity.configure({
...
+  },
+  migrate(item) {
+    item.description = '';
+    item.stability = 'experimental';
+    item.documentation = '';
+    item.payloadSchema = null;
```
jonasfj: Null validated against nothing...

You'll want `true`, it's the simplest JSON schema... See schema spec..

src/data.js
jonasfj: Or maybe we limit it to 64kb

src/data.js
helfi92: Done

src/data.js
helfi92: Do I also need to change [L97](https://github.com/helfi92/taskcluster-queue/blob/2e915b7597dd88454b1c0c1e40604219c856bdb3/src/workerinfo.js#L97) to true? I'm only asking because when I do that, the tests are failing most probably because the schema was expecting an object/null but received a boolean value.

schemas/list-workertypes-response.yml
djmitche: Per IRC discussion, let's drop documentation and schema entirely, rather than continue to worry about them here.  We can come back in a later PR and add them, hopefully only in the get-workertype-response, and possibly in an azure blob.

Description and stability and expires can stay.  `provisionerId`, also added here, seems a little redundant but it doesn't hurt.

schemas/list-workertypes-response.yml
```diff
@@ -20,10 +20,45 @@ properties:
...
+          title:        "Documentation"
+          description: |
+            Documentation for the worker-type.
+          type:         string
+        expires:
+          title:        "Worker-type Expiration"
+          description: |
+            Date and time after which the worker-type created will be automatically
```
djmitche: s/created //

helfi92:
@djmitche OK, `documentation` and `payloadSchema` are now omitted :)

PR #189, opened 2017-08-24
Bug 1390866 - Implement API endpoint to GET and PUT a provisioner

helfi92:


schemas/provisioner-response.yml
```diff
@@ -0,0 +1,48 @@
+$schema:  http://json-schema.org/draft-04/schema#
+title:          "Provisioner Response"
+description: |
+  Response to an attempt to request data from a provisioner.
```
djmitche: "Response containing information about a provisioner"

src/api.js
```diff
@@ -2016,6 +2016,97 @@ api.declare({
...
+api.declare({
+  method:     'put',
+  route:      '/provisioners/:provisionerId',
+  name:       'declareProvisioner',
+  stability:  API.stability.experimental,
+  scopes:     [
+    [
+      'queue:declare-provisioner:<provisionerId>#<property>',
+    ],
+  ],
```
djmitche: needs `deferAuth` too

helfi92: I thought I had added it. Good catch!

helfi92:
@djmitche Will merge in ~ half an hour if no complaints :)

PR #190, opened 2017-08-25
Implement API endpoint to GET and PUT a worker-type

helfi92:


schemas/workertype-response.yml
```diff
@@ -0,0 +1,55 @@
...
+    title:        "Worker-type Expiration"
+    description: |
+      Date and time after which the worker-type will be automatically
+      deleted by the queue.
+    type:         string
+    format:       date-time
+  lastDateActive:
+    title:        "Worker-type Last Date Active"
+    description: |
+      Date and time where the worker-type was last seen active
```
djmitche: Let's add a note here similar to that for the lastDateUsed for clientIds in the auth API.  That says "Will only be updated every 6 hours or so this may be off by up-to 6 hours. But it still gives a solid hint as to whether or not this client is in use."  The grammar could be improved :)

src/api.js
```diff
@@ -2087,6 +2087,79 @@ api.declare({
...
+
+  const wType = await this.WorkerType.scan({
+    provisionerId,
+    workerType,
+    expires: Entity.op.greaterThan(new Date()),
+  });
+
+  const result = wType.entries.length ? wType.entries[0].json() : {};
+
+  return wType.entries.length ?
```
djmitche: Rather than two `?` expressions, a conditional with `return res.reportError(..)` would be simpler.

src/api.js
```diff
@@ -2087,6 +2087,79 @@ api.declare({
...
+    res.reply(result) :
+    res.reportError('ResourceNotFound',
+      'Worker-type {{workerType}} with  Provisioner {{provisionerId}} not found. Are you sure it was created?', {
```
djmitche: two spaces

src/api.js
```diff
@@ -2087,6 +2087,79 @@ api.declare({
...
+  description: [
+    'Update a worker-type from a provisioner.',
+  ].join('\n'),
+}, async function(req, res) {
+  const {provisionerId, workerType} = req.params;
+  const {stability, description, expires} = req.body;
+
+  const wType = await this.WorkerType.load({provisionerId, workerType}, true);
+
+  if (!wType) {
```
djmitche: Per the agreement to be permissive, if the entity doesn't exist, then this call should create it.

src/api.js
```diff
@@ -2087,6 +2087,79 @@ api.declare({
...
+    res.reportError('ResourceNotFound',
+      'Worker-type {{workerType}} with  Provisioner {{provisionerId}} not found. Are you sure it was created?', {
+        workerType,
+        provisionerId,
+      },
+    );
+  }
+
+  const result = await wType.modify((entity) => {
```
djmitche: This will need some scopes checked first!  I liked Jonas's suggestion:
```
scopes:    'queue:declare-worker-type:<provisionerId>/<workerType>#<property>'
```
with the property checked for each property included in the request.  Then we can control who can update which property.

helfi92: I declared `scopes` in the options: https://github.com/taskcluster/taskcluster-queue/pull/190/commits/9a59d3bc5dccc665562b989d86430e49c0a5ab80#diff-e1d62d8c4965edf0e532dfa19da0891fR2128 wouldn't that do it?

Also, just out of curiosity, could you give me an example of `<property>`

djmitche: Oh, I did miss that.  But we need more, since the framework isn't going to know what to do with `<property>`.  You'll need to add the `deferAuth` option and call req.satisfies a few times, once for each property.  `<property>` in this case would be `stability`, `description`, or `expires`.

helfi92: Makes sense, thanks for the clarifications.

src/api.js
```diff
@@ -2087,6 +2087,91 @@ api.declare({
...
+  stability:  API.stability.experimental,
+  output:     'workertype-response.json#',
+  title:      'Get a worker-type',
+  description: [
+    'Get a worker-type from a provisioner.',
+  ].join('\n'),
+}, async function(req, res) {
+  const {provisionerId, workerType} = req.params;
+
+  const wType = await this.WorkerType.scan({
```
djmitche: @jonasfj is scan the right choice here?

src/api.js
```diff
@@ -2087,6 +2087,91 @@ api.declare({
...
+    provisionerId,
+    workerType,
+    expires: Entity.op.greaterThan(new Date()),
+  });
+
+  const result = wType.entries.length ? wType.entries[0].json() : {};
```
djmitche: you can just move this line down and remove the `?`

src/api.js
```diff
@@ -2087,6 +2087,91 @@ api.declare({
...
+}, async function(req, res) {
+  const {provisionerId, workerType} = req.params;
+  const {stability, description, expires} = req.body;
+
+  const wType = await this.WorkerType.load({provisionerId, workerType}, true);
+
+  // Authenticate request by providing parameters
+  const requestAllowed = req.satisfies({provisionerId, workerType, property: stability}) &&
+    req.satisfies({provisionerId, workerType, property: description}) &&
+    req.satisfies({provisionerId, workerType, property: expires});
```
djmitche: These would be `property: 'stability'`, `property: 'description'`, etc.  That is, the completed scopes are things like `queue:declare-worker-type:aws-provisioner-v1/gecko-b-1-w2008#expires`.

Also, if the property isn't provided in the request, we shouldn't check for the scope.  So a request with `{description: 'This worker type is great'}` would only require `queue:declare-worker-type:aws-provisioner-v1/gecko-b-1-w2008#desscription`.  But if the request also included expires, the `#expires` scope would be required too.

djmitche: This probably should have a few sentences in the API description to explain it, too.

src/api.js
helfi92: Good catch. Since we're only returning at most 1 worker-type, this should probably use `load`.

src/api.js
helfi92: That was a stupid mistake from my part. Thanks!

src/api.js
```diff
@@ -2087,6 +2087,93 @@ api.declare({
...
+    [
+      'queue:declare-worker-type:<provisionerId>/<workerType>#<property>',
+    ],
+  ],
+  deferAuth:  true,
+  output:     'workertype-response.json#',
+  input:      'update-workertype-request.json#',
+  title:      'Update a worker-type',
+  description: [
+    'Update a worker-type from a provisioner.',
```
djmitche: "Declare a workerType, supplying some details about it"

src/api.js
```diff
@@ -2087,6 +2087,93 @@ api.declare({
...
+      entity.description = description || entity.description;
+      entity.expires = new Date(expires || entity.expires);
+    }) :
+    await this.WorkerType.create({
+      provisionerId,
+      workerType,
+      expires,
+      lastDateActive: new Date(),
+      description,
+      stability,
```
djmitche: I think expires, description, and stability need default values for the create case.

src/workerinfo.js
```diff
@@ -2,20 +2,29 @@ const taskcluster = require('taskcluster-client');
...
+
 const ignoreEntityAlreadyExists = (err) => {
   if (!err || err.code !== 'EntityAlreadyExists') {
     throw err;
   }
 };
 
-const expired = entity => new Date(entity.expires) - Date.now() < 24 * 60 * 60 * 1000;
+const expired = entity => new Date(entity.expires) - Date.now() < DAY;
+const shouldUpdateLastDateActive = entity => Date.now() - new Date(entity.lastDateActive) > DAY / 4;
```
djmitche: let's do the subtraction and `<` in the same direction for both :)

helfi92: I think this way is more human readable as it makes more sense. What did you have in mind?

djmitche: I don't mind which way, but my complaint is that one is "somedate - now < something" while the other is "now - somedate > something".  So I had to stare at it a while to figure out that they're both true after somedate + something.  How about

```js
const expired = entity => Date.now() > new Date(entity.expires) + DAY;
const shouldUpdateLastDateActive = entity => Date.now() > new Date(entity.lastDateActive) +  DAY / 4; 
```

helfi92: @djmitche The `expired` suggestion is saying to set `expired` to `true` only when we have gone 1 day past the expiration date. Wouldn't the ideal case be to set `expired` to true if we are within 1 day of the expiration?

djmitche: You're right.  I flipped the sign on the inequality correctly.  See, it is confusing!

```js
const expired = entity => Date.now() > new Date(entity.expires) - DAY;
```

helfi92:
Added the last final tweaks. I will go ahead and merge in about half an hour if nobody complains :)

djmitche:
+1

PR #191, opened 2017-08-29
Blob artifacts aug 29

jhford:
Checking on artifacts being present before resolving task is outside the scope of this work and needs to be done in a follow up.


jhford:
I cannot run the tests locally at all any longer, so I have no clue if there are any errors with the added code

jonasfj:
You have test failures:

```
  2) Artifacts Blob Storage Type S3 multi part complete flow:
     Error: Internal Server Error, incidentId: 004d4afd-7f9a-4a05-9118-53df5349e2f7
----
method:     completeArtifact
errorCode:  InternalServerError
statusCode: 500
time:       2017-08-31T08:04:35.043Z
      at node_modules/taskcluster-client/lib/client.js:327:21
      at tryCallOne (node_modules/promise/lib/core.js:37:12)
      at node_modules/promise/lib/core.js:123:15
      at flush (node_modules/asap/raw.js:50:29)
      at _combinedTickCallback (internal/process/next_tick.js:131:7)
      at process._tickDomainCallback (internal/process/next_tick.js:218:9)
  3) Artifacts Blob Storage Type S3 multi part idempotency:
     Unknown assertation failure occured, assumed `'20170831T080436Z'` to not equal (===) `'20170831T080436`
```

Note: tests on master runs perfect on my laptop, you can ask hassan, but afaik he has it running locally too (as he has been merging things :) )

jhford:
@jonasfj the tests all run through completion on my machine.  Can you please review the code regardless of this failure, I'm sure it's a small fix to get the test to work?

schemas/put-artifact-response.yml
```diff
@@ -0,0 +1,13 @@
+$schema:            http://json-schema.org/draft-04/schema#
+title:              "Put Artifact Response"
+description: complete a blob artifact upload
+type:               object
+properties:
+  outcome:
+    type: string
+    enum:
+      - success
+      - failure
```
jonasfj: There is no response from this end-point... If you really want one, return the JSON from the `artifact.json()` or whatever it is we report in `listArtifacts`

src/api.js
```diff
@@ -1678,6 +1684,26 @@ var resolveTask = async function(req, res, taskId, runId, target) {
...
+      runId,
+      storageType: 'blob',
+      present: false,
+    }, {
+      limit: 1,
+      handler: () => { haveAllBlobs = false; },
+    });
+
+    if (!haveAllBlobs) {
+      return res.reportError('InputError',
```
jonasfj: Or  `RequestConflict` as it's sort of the request that conflicts with the state...

jonasfj: btw, thanks for adding this... It's a huge safety feature... and means we can trust the workers a little bit less :)

src/artifacts.js
```diff
@@ -183,8 +195,70 @@ api.declare({
...
+      // Instead of storing the entire parts list in the entity, we'll instead
+      // store a hash of the JSON serialized version of the list.
+      //
+      // Note that this means that for loaded entities we'll need to use the
+      // details.partsHash field to check whether we're multipart or not
+      // instead of the details.parts list
+      if (input.parts) {
+        let partsHash = crypto.createHash('sha256');
+        for (let part of input.parts) {
+          partsHash.update(`${part.sha256}${part.size}`);
```
jonasfj: May I suggest `${part.sha256}:${part.size}\n`, I know it's nit... because json schema ensures you can't provide these of arbitrary lengths... But it's a nice extra safety feature..

src/artifacts.js
```diff
@@ -248,8 +329,42 @@ api.declare({
...
+      // attribute.  In that case, we'll cancel the uploadId we just created
+      if (_.isEqual(artifact.details, details)) {
+        // Do nothing!
+      } else if (input.parts && _.isEqual(storedWithoutUploadId, inputWithoutUploadId)) {
+        if (artifact.details.uploadId !== details.uploadId) {
+          await this.s3Controller.abortMultipartUpload({
+            bucket: details.bucket,
+            key: details.key,
+            uploadId: details.uploadId,
+          });
```
jonasfj: Don't we also have to set `details.uploadId = artifact.details.uploadId`? That would explain the error wrt. idempotency we saw in tests in travis...

I think we have `modify(...)` which sets `artifact.details  = details` later...

jonasfj: Another option is `details = artifact.details` that's probably safer...

Note: I could be wrong here, maybe I misunderstood the flow...

src/artifacts.js
```diff
@@ -423,6 +643,119 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+  let input       = req.body;
+  let isPublic    = /^public\//.test(name);
+
+  // Hmm, list destructuring is probably better here
+  let [artifact, task] = await Promise.all([
+    this.Artifact.load({taskId, runId, name}, true),
+    this.Task.load({taskId}, true),
+  ]);
+
+  // TODO: Have a shortcut here where things which are already
```
jonasfj: That's just adding `if (artifact.present) { return res.status(200).json({});}`, maybe add it after checking that the run exists :)

src/artifacts.js
```diff
@@ -423,6 +643,119 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+      artifact: artifact.json(),
+      workerGroup,
+      workerId,
+      runId,
+    }, task.routes);
+    return res.reply({
+      outcome: 'success',
+    });
+  } else {
+    throw new Error('Only supported storageType artifacts may be marked completed');
```
jonasfj: 1) Always return early as to simplify the code
2) This generates an internal error... that is a bug (by definition).

Instead do `if (artifact.storageType != 'blob') { return res.reportError('RequestConflict', ....);}

src/artifacts.js
```diff
@@ -423,6 +643,119 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+
+    await this.publisher.artifactCreated({
+      status: task.status(),
+      artifact: artifact.json(),
+      workerGroup,
+      workerId,
+      runId,
+    }, task.routes);
+    return res.reply({
+      outcome: 'success',
```
jonasfj: Just do: `return res.status(200).json({})`, we don't have any content here...

jonasfj: Or maybe do: `return res.reply({artifact: artifact.json()})` that's a somewhat useful thing to respond and consistent with what we return in listing of artifacts :)

src/artifacts.js
```diff
@@ -423,6 +643,119 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+    }
+
+    await artifact.modify((artifact) => {
+      artifact.details.etag = etag;
+      // Now that we're finished, we don't want to store the uploadId any longer
+      artifact.details = _.omit(artifact.details, 'uploadId');
+      artifact.present = true;
+    });
+
+    await this.publisher.artifactCreated({
```
jonasfj: Publish before you modify... If publication fails and client retries (as required), the next request won't publish again, because you'll want to return early when `artifact.present == true`

src/artifacts.js
```diff
@@ -423,6 +643,119 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+    workerId,
+    name,
+  })) {
+    return;
+  }
+
+  if (artifact.storageType === 'blob') {
+    let etag;
+    if (artifact.details.partsHash) {
+      etag = await this.s3Controller.completeMultipartUpload({
```
jonasfj: Do we have any tests that we can call this twice... or do we need to do a `HEAD` request if this fails? inorder to ensure things will work in case of idempotency?

src/artifacts.js
```diff
@@ -423,6 +643,119 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+      let headRes = await this.s3Runner.run({
+        req:{
+          url,
+          method: 'HEAD',
+          headers: {},
+        },
+      });
+      
+      if (headRes.headers['x-amz-meta-content-sha256'] !== artifact.details.contentSha256) {
+        throw new Error('S3 object does not have correct Content-Sha256 metadata');
```
jonasfj: Maybe add a comment explaining why this should not be possible... Perhaps even insert a label into the error indicating that this could be a sign of an attack... right? I mean, this would sort of only be possible if someone was actively trying to insert invalid artifacts...

src/artifacts.js
```diff
@@ -423,6 +643,119 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+      }
+      
+      etag = input.etags[0];
```
jonasfj: Shouldn't we test this against the `etag` returns from the S3 HEAD request? and report InputError if the two don't match?

schemas/put-artifact-response.yml
jhford: ok, removed schema and stopped sending an output.  This will now send 204 on success

src/api.js
jhford: no worries.

src/artifacts.js
jhford: we could do that, but this string is never shown anywhere and is just used to generate a sha256 sum.  First, I'd rather not use `\n` because that maps to different characters on different systems.  Second, we're never showing this string anywhere, so the actual format is completely unimportant, so having the simplest format is fine.

I'll change the string to `${part.sha256}:${part.size}_` so that in the off chance we did want to display the string, we can without adding complexity to the string generation.

src/artifacts.js
jhford: No, we don't store the `uploadId` we just created, we cancel the one that we just created and keep the old one and use the old one.

jhford: I'm not sure what is causing the failure on travis, because it runs perfectly on my mac locally.  I need to debug the test a little more

src/artifacts.js
jhford: the number of times we'll hit this is very small, so let's leave the full codepath in place.

src/artifacts.js
jhford: I've added some tests for that.

src/artifacts.js
jhford: I don't know that it strictly implies an attack, rather it just shows that someone is uploading incorrectly.  What would be a stronger indication of an attack is that the header and artifact.details.contentSha256 values are the same, but the sha256 of the artifact as stored in S3 doesn't match either.

This is more of a sanity check and doesn't give us the strong check that signing the requests to only work with a specific payload body does.

src/artifacts.js
jhford: That assumes eventual consistency is working as expected, and that no one has done an intermediate GET/HEAD before our HEAD which could cause a badly-cached resource to be stored in the S3 system.

src/artifacts.js
jhford: What if publication fails 5 times and the retry limit is 5, that means that we'd send a message that the artifact is created, but we wouldn't have modified the artifact so the values in the entity would be all wrong, and we wouldn't hit the shortcircuiting logic if it is there.

That seems like a bad idea to me.

src/artifacts.js
jhford: I've changed to res.reportError, but this is as early a return as should be made, even if it doesn't visually appear that way.  We could return before doing the scope checking, but that could leak information about the existence of private artifacts.

jhford:
Closing this PR to open a new one.

src/artifacts.js
jonasfj: `\n` is NOT platform dependent... `\n` is LF (line feed), `\r` is CR (carriage return).
https://en.wikipedia.org/wiki/Newline

`\n` is not newline (yes, newline is platform dependent), but you'll rarely see anyone use anything but LF for internal strings...

src/artifacts.js
jonasfj: Correct, but we do make a `.modify` this would cause the `.modify` to be a  no-op.

src/artifacts.js
jonasfj: Is `x-amz-meta-content-sha256` covered by the signature?

If **no** then this is a 4xx error, because the user changed headers from the upload request.

if **yes** then a request upload this file should not be possible, the queue should never have signed it. Hence, someone changed the database, someone has S3 access, or we have a massive bug. Eitherway, it's serious as in we're under attack or have a massive flaw.

jonasfj: @jhford, note: you can't throw an error if this isn't an internal error, if it can be provoked by the client throwing an error is a bug, as it results in internal error.

src/artifacts.js
jonasfj: > That assumes eventual consistency is working as expected

No eventual consistency involved here, we rely on read-after-write consistency for first PUT.

> that no one has done an intermediate GET/HEAD before our HEAD which could cause a badly-cached resource to be stored in the S3 system.

Elaborate? I don't see how anything like that can change the etag.

Plausibly, if you had two PUT requests where the last one changes the etag.
Afaik, etags are content based and can't be changed by uploading the same thing again, am I wrong here?
Also we sign everything in the headers for the requests so two different resources cannot be uploaded.

Please outline the cases in which the etag can be made inconsistent by worker...

jonasfj: @jhford ^

src/artifacts.js
jonasfj: @jhford, I have an email that says:
> What if publication fails 5 times and the retry limit is 5, that means that we'd send a message that the artifact is created, but we wouldn't have modified the artifact so the values in the entity would be all wrong, and we wouldn't hit the shortcircuiting logic if it is there.
That seems like a bad idea to me.

You have a point that we return `404` until `present = true` so we don't want to send a message that the artifact is done until such time that `present = true`... Fair argument...

You can fix this by changing the comment that says `// TODO: Return early if present = true` to say that we **cannot** return early, without sending the `artifactCreated` message again...
Either make the code, or make the comment very clear...Otherwise we can't ensure that messages arrive...

src/artifacts.js
jonasfj: @jhford,

Don't do:
```js
if (artifact.storageType === 'blob') {
  // bla bla...
} else {
  throw Error('...');
}
```
Instead do:
```js
if (artifact.storageType !== 'blob') {
  throw Error('...');
}

// bla bla...
```
Hence, **prefer to return early**..

Also, and I can't stress this enough... do not throw an error incase of input error. **input error => 4xx** not 5xx

src/artifacts.js
```diff
@@ -423,6 +643,117 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+
+  let run = task.runs[runId];
+  if (!run) {
+    return res.reportError('InputError',
+      'Run not found',
+      {});
+  }
+  let workerGroup = run.workerGroup;
+  let workerId = run.workerId;
+
```
jonasfj: @jhford, we should return 409, `RequestConflict` if the run have been resolved...

Specifically, if `run.state` is not `running` or `exception` (with `run.resolved` within the past 25 minutes).
Same condition we have for creating artifacts:
https://github.com/taskcluster/taskcluster-queue/blob/9b36857aa8f8f27515382b48e638e9754ebf5993/src/artifacts.js#L160-L181

PR #194, opened 2017-09-01
Bug 1390870 - Implement API endpoint for a worker

helfi92:
API endpoint to `GET` and `PUT` a worker

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -9,7 +9,7 @@ let Promise     = require('promise');
 
 // Maximum number runs allowed
 const MAX_RUNS_ALLOWED = 50;
-
+const SLUGID_SIZE = 128 / 8;
 // Priority levels in order from high to low
 const PRIORITY_LEVELS = [
   'highest',
@@ -1373,6 +1373,8 @@ api.declare({
     this.workerInfo.seen(provisionerId, workerType, workerGroup, workerId),
   ]);
 
+  await this.workerInfo.taskSeen(provisionerId, workerType, workerGroup, workerId, result);
+
   return res.reply({
     tasks: result,
   });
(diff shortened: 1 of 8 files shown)


src/api.js
```diff
@@ -9,7 +9,7 @@ let Promise     = require('promise');
 
 // Maximum number runs allowed
 const MAX_RUNS_ALLOWED = 50;
-
+const SLUGID_SIZE = 128 / 8;
```
djmitche: ??

src/api.js
```diff
@@ -1373,6 +1373,8 @@ api.declare({
     this.workerInfo.seen(provisionerId, workerType, workerGroup, workerId),
   ]);
 
+  await this.workerInfo.taskSeen(provisionerId, workerType, workerGroup, workerId, result);
```
djmitche: This should get done in the same Promise.all above, so that it occurs in parallel.

djmitche: Although that leaves a bit of a race between workerInfo.seen creating a worker entity and workerInfo.taskSeen updating its list of tasks.  Could this all be included in workerInfo.seen, by passing `tasks` to it?

src/api.js
helfi92: Oops, unused constant. Will remove.

src/api.js
helfi92: @djmitche Hm, tasks (`result`) is only available after `this.workClaimer.claim` resolves and so if this happens in parallel to workerInfo.seen, I won't be able to pass `tasks` to it.

djmitche: Oh, I didn't see that.  OK, that solves the race, too.

PR #196, opened 2017-09-06
Bug 1397373 - describe superseding in queue reference

djmitche:
(mostly copied from docker-worker docs)

### DIFF
--- a/docs/superseding.md
+++ b/docs/superseding.md
@@ -1,0 +1,56 @@
+---
+title: Superseding
+---
+
+In many cases, executing some tasks in the queue can render others unnecessary.
+For example, if the test tasks for a later version-control revision succeed, then tests for earlier revisions can be skipped.
+Such earlier tasks are said to be "superseded" by the later task.
+
+The queue supports superseding by implementing a `superseded` reason for task resolution (seen with state `exception`).
+The rest is up to the worker.
+
+## Convention
+
+General-purpose workers that support superseding are encouraged to follow this convention, allowing users of those workers to use common code to generate superseding configuratoin.
+Note that not all workers support this convention, and that some special-purpose workers may implement something completely different.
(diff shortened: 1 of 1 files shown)


docs/superseding.md
```diff
@@ -0,0 +1,48 @@
...
+---
+
+In many cases, executing some tasks in the queue can render others unnecessary.
+For example, if the test tasks for a later version-control revision succeed, then tests for earlier revisions can be skipped.
+Such earlier tasks are said to be "superseded" by the later task.
+
+Taskcluster defines a convention for superseding multiple tasks, executing only one (the "primary task").
+Crucially, the primary task need not be the same task that the worker found on the queue, but may be one deeper into the queue (e.g., a change pushed to version control later).
+
+Superseding is implemented by workers, and is only supported by a few worker implementations -- consult their documentation for details.
```
petemoore: minor nit: in the taskcluster-docs site, we said _most_ workers support superseding, and here we say _only a few_ - we should probably align the two. Maybe we should say "not all" and provide a link for the curious reader that wants to know (and suggest to the user to browse worker docs to find out if a particular worker implements the feature or not).

docs/superseding.md
```diff
@@ -0,0 +1,48 @@
...
+
+Example:
+
+```js
+{
+    "supersederUrl": "https://foo-coalescer.herokuapp.com/build/linux/master"
+}
+```
+
+The worker appends the taskId of the task it has received from the queue (the "initial taskId") as a query argument, `?taskId=<taskId>`.
```
jonasfj:
```md
`taskId`
```

djmitche:
I updated this to be explicit about calling it a convention.

I think that TC is at the point in its evolution where it makes sense to define some conventions.  We really don't have a place for that in the docs right now, so maybe we should add sometihng -- but that's another conversation for another time.

PR #197, opened 2017-09-11
Blob artifacts sep 11

jhford:
If this passes travis tests, we should be good to land.  Otherwise, I cannot duplicate the travis failures locally, so I need a PR with which to test.


jhford:
I'm not sure why this is failing on travis, but it seems to be related to permissions over anything wrong with the code. @jonasfj are you able to run this branch on your machine?

jhford:
Locally, I get `  191 passing (7m)` for this branch.  Maybe the issue is the IAM profile in use?

jonasfj:
@jhford, I added some comments to the last PR in response to some of the resolutions please see them...

src/artifacts.js
```diff
@@ -423,6 +643,117 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+
+  let run = task.runs[runId];
+  if (!run) {
+    return res.reportError('InputError',
+      'Run not found',
+      {});
+  }
+  let workerGroup = run.workerGroup;
+  let workerId = run.workerId;
+
```
jonasfj: Need condition on `run.state`, see: https://github.com/taskcluster/taskcluster-queue/blob/9b36857aa8f8f27515382b48e638e9754ebf5993/src/artifacts.js#L160-L181

src/artifacts.js
```diff
@@ -423,6 +643,117 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+    await this.publisher.artifactCreated({
+      status: task.status(),
+      artifact: artifact.json(),
+      workerGroup,
+      workerId,
+      runId,
+    }, task.routes);
+    return res.status(204).send();
+  } else {
+    return res.reportError('InputError',
```
jonasfj: return early

src/artifacts.js
```diff
@@ -423,6 +643,117 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+    workerId,
+    name,
+  })) {
+    return;
+  }
+
+  if (artifact.storageType === 'blob') {
+    // If the artifact is present, we've already done what's required here
+    if (artifact.present) {
+      return res.status(204).send();
```
jonasfj: Must publish message before doing this!

jonasfj:
@jhford, I don't really have much to add.. If I'm still jetlagged tomorrow I'll ping you early and we can chat about what's left... Maybe we can scan the code while chatting, just root out the rest...

I'm starting to feel like we should be ready to ship this soon.. :)

PR #198, opened 2017-09-12
Bug 1399123 - Enable/Disable workers

helfi92:


src/api.js
```diff
@@ -2391,6 +2391,7 @@ api.declare({
   const result = worker ?
```
jonasfj: probably don't do a ternary like this, it's a bit hard to read... neat, but hard to read..

src/api.js
```diff
@@ -2399,6 +2400,7 @@ api.declare({
       workerId,
       recentTasks: Entity.types.SlugIdArray.create(),
       expires: expires || taskcluster.fromNow('1 day'),
+      disabled: false,
       firstClaim: new Date(),
     });
```
jonasfj: You'll need to catch the case there the entity exists... in case you are racing...

Note: I might be tempting to make a `this.Worker.upsert` method in `data.js` that does the right thing in these cases... Assuming it can be easily defined what that is for the `Worker` table..

src/api.js
helfi92: Hehe, ok. Changing this one and the ones above it to use `if` statements

src/api.js
helfi92: Sounds good

src/workerinfo.js
```diff
@@ -174,7 +175,7 @@ class WorkerInfo {
       workerId,
     }, true);
 
-    if (!tasks.length || !worker) {
+    if (!tasks.length || !worker || worker.disabled) {
```
djmitche: It would be a bug if taskSeen was called with a disabled worker -- but we should probably record the taskId in that case.  Otherwise we'd be hiding a bug.

helfi92: Hm, not sure I follow. Presently, a worker can be disabled as long as the worker has not expired. Once expired, a new one will be created. If one wants to make sure a worker doesn't expire, then they could simply change the expiration date of the worker through `declareWorker`. That said, why would it be a bug if `taskSeen` was called with a disabled worker? One of the first thing this method does is check if the worker is disabled. If it is, then it returns without recording the task ID as seen.

jonasfj:
I have to ask where do you check if a worker is disabled before assign it work in claimWork and claimTask?
(also maybe we should load test that, as it adds an additional entity load per request)

src/workerinfo.js
djmitche: It would be a bug because it would mean that a task was assigned to a disabled worker..

helfi92:
Spoke to @jonasfj and he seems to be happy with this change. @djmitche I'm planning to merge tomorrow morning unless I hear complaints :)

djmitche:
Jonas is smarter than me, so that's good enough for me!

PR #199, opened 2017-09-12
Count claimWork calls

jonasfj:
We can use this to create alerts in signalfx, if some workerType isn't claiming work anymore...


helfi92:
@jonasfj Weird, Travis is failing.

djmitche:
@jonasfj or @helfi92 should we still land this?

PR #200, opened 2017-09-14
Patch to fix sentry error

helfi92:


src/api.js
```diff
@@ -1479,7 +1479,7 @@ api.declare({
     this.workerInfo.seen(task.provisionerId),
   ]);
 
-  await this.workerInfo.taskSeen(task.provisionerId, task.workerType, workerGroup, workerId, result);
+  await this.workerInfo.taskSeen(task.provisionerId, task.workerType, workerGroup, workerId, [result]);
```
jonasfj: You need to read the docs for `WorkClaimer.claimTask`, see:
https://github.com/taskcluster/taskcluster-queue/blob/master/src/workclaimer.js#L203-L208

It might not return a task claim at all...

src/api.js
```diff
@@ -1502,6 +1500,10 @@ api.declare({
     );
   }
 
+  await this.workerInfo.taskSeen(
+    task.provisionerId, task.workerType, workerGroup, workerId, typeof result === 'string' ? [] : [result]
```
jonasfj: No need for the ternary anymore...

PR #201, opened 2017-09-19
Blob artifacts sep 19

jhford:
I think this should be ready to go.  I'm not sure what is up with travis failures, since I can run these tests completely fine locally.  My suspicion is that it's related to the S3 user account being limited to not have the required permissions or something like that.


jonasfj:
TODO: Create production bucket

gregarndt:
I have never been happier to see two words! \o/

PR #204, opened 2017-09-20
Bug 1401360 - Modify taskSeen to use the new SlugIdArrray methods

helfi92:


src/workerinfo.js
```diff
@@ -183,35 +183,27 @@ class WorkerInfo {
...
+        recentTasks.shift();
+      }
     });
+
+    try {
+      await worker.modify(entity => {
+        entity.recentTasks = recentTasks;
+      });
+    } catch (err) {
+      throw err;
```
jonasfj: huh?

PR #205, opened 2017-09-21
Bug 1401961 - Extend listWorkers endpoint to allow filtering by disabled workers

helfi92:
For the page listing workers, one would want to filter results. Since we are dealing with a possibly long list, it's hard to filter when you have a paginated list. To solve this issue, we should return a filtered list instead. This pull-request adds `disabled` to the list of options.

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -2308,6 +2308,7 @@ api.declare({
   query: {
     continuationToken: /./,
     limit: /^[0-9]+$/,
+    disabled: /^(true|false)$/,
   },
   name:       'listWorkers',
   stability:  API.stability.experimental,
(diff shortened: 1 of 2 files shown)


src/api.js
```diff
@@ -2323,15 +2324,22 @@ api.declare({
   ].join('\n'),
 }, async function(req, res) {
   const continuation = req.query.continuationToken || null;
+  const disabled = req.query.disabled || null;
   const provisionerId = req.params.provisionerId;
   const workerType = req.params.workerType;
   const limit = Math.min(1000, parseInt(req.query.limit || 1000, 10));
 
-  const workers = await this.Worker.scan({
+  const worker = {
```
djmitche: Maybe `workerQuery`?  This confused me for a minute..

src/api.js
```diff
@@ -2308,6 +2308,7 @@ api.declare({
   query: {
     continuationToken: /./,
     limit: /^[0-9]+$/,
+    disabled: /^(true|false)$/,
```
djmitche: Add a sentence or two to the method doc for this, too, please :)

jonasfj:
nice! :)

PR #206, opened 2017-09-21
Blob artifacts -- just for unit tests.

jhford:


jhford:
I'm going to use my admin privs to do this, but I have r+ in #201

PR #210, opened 2017-09-30
Bug 1404010 - Extend the worker to have the runId of its most recentTask

helfi92:
Use `Entity.types.JSON` with `[{taskId, runId}, ...]` instead of `Entity.types.SlugIdArray`.

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -912,9 +912,7 @@ let Worker = Entity.configure({
     firstClaim:       Entity.types.Date,
   },
   migrate(item) {
-    item.firstClaim = new Date(2000, 0, 1);
     item.recentTasks = [];
-    item.disabled = false;
 
     return item;
   },
--- a/src/data.js
+++ b/src/data.js
@@ -892,6 +892,32 @@ let Worker = Entity.configure({
 
     return item;
   },
+}).configure({
+  version:            4,
+  properties: {
+    provisionerId:    Entity.types.String,
+    workerType:       Entity.types.String,
+    workerGroup:      Entity.types.String,
+    workerId:         Entity.types.String,
+    /**
+     * List of objects with properties:
+     * - taskId
+     * - runId
+     * See JSON schema for documentation.
+     */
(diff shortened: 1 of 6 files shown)


src/data.js
```diff
@@ -892,6 +892,32 @@ let Worker = Entity.configure({
...
+    recentTasks:      Entity.types.JSON,
+    disabled:         Entity.types.Boolean,
+    // the time at which this worker should no longer be displayed
+    expires:          Entity.types.Date,
+    firstClaim:       Entity.types.Date,
+  },
+  migrate(item) {
+    item.firstClaim = new Date(2000, 0, 1);
+    item.recentTasks = [];
+    item.disabled = false;
```
djmitche: Only `recentChanges` needs to be updated here -- the other two fields have not changed.

src/workerinfo.js
```diff
@@ -183,13 +182,14 @@ class WorkerInfo {
...
     }
 
-    const recentTasks = worker.recentTasks.clone();
+    const recentTasks = [...worker.recentTasks];
 
-    tasks.forEach((task, index) => {
-      const taskId = tasks[index].status.taskId;
+    tasks.forEach((task) => {
+      const {taskId, runs} = task.status;
+      const runId = runs.length;
```
djmitche: This will start at 1 for runId 0, won't it?

djmitche: (I'm thinking yes given all the runId: 1 in the tests..)

helfi92: Good catch!

helfi92:
@jonasfj I wrote it like the following:

```js
await worker.modify(worker => {
  worker.recentTasks = [
    ...worker.recentTasks,
    ...tasks.map(({status}) => ({taskId: status.taskId, runId: status.runs.length - 1})),
  ].slice(-RECENT_TASKS_LIMIT);
});
```

That way, we make sure it always has the most recent tasks showing in case `tasks` has length greater than `RECENT_TASKS_LIMIT` (assuming this is possible...not sure).

PR #212, opened 2017-10-16
Bug 1409186 - Implement actions on the Provisioner context

helfi92:


helfi92:
I'll have a different PR with regards to actions on a WorkerType and Worker.

schemas/list-provisioners-response.yml
```diff
@@ -41,6 +41,64 @@ properties:
...
           format:       date-time
+        actions:
+          type:         array
+          items:
+            type:       object
+            title:      Actions
+            description: |
+              Actions can be defined at any context higher than or at the
+              action's level -- so a worker type or provisioner can have
+              actions with level=worker. This should avoid unnecessary
```
jonasfj: `level` was renamed `context`

schemas/list-provisioners-response.yml
```diff
@@ -41,6 +41,64 @@ properties:
...
+                  When the action is triggered, it will make a POST request to this URL.
+                  The request will be signed with the user's Taskcluster credentials.
+                type:         string
+                format:       uri
+              description:
+                title:        "Description"
+                description: |
+                  Description of the provisioner.
+                type:         string
+              severity:
```
jonasfj: I propose that we do not implement this...
Instead we make special handling of certain labels and document this by convention.
I could see the tools site mapping a fixed set of labels to icons, and using a generic icon for actions with other labels.
This is also in-line with how `actions.json` works.

It'll be best effort, but that's IMO fine for this purpose.. Introducing a concept we probably don't need is overkill, if we really need it we can do it later..

note: all these actions should require confirmation anyways.

schemas/provisioner-response.yml
```diff
@@ -39,10 +39,69 @@ properties:
...
+          type:         string
+          enum:         ["safe", "warning", "danger"]
+      additionalProperties: false
+      required:
+        - label
+        - title
+        - context
+        - url
+        - description
+        - severity
```
jonasfj: In the return value `actions` should be required...

Ie. when you `declareProvisioner` the `actions` property is optional (default to `[]`), when you `listProvisioners` we return  the default value `[]` if no actions was specified..

This way api consumers can always rely on the property being present, and they can always assume it's a list (even if it's just an empty list). Which makes it a lot easier to build stuff using it.
It's a concept we follow in pretty much all other APIs, and it's generally nice not have `.. || []` all over your code in tc-tools :)

helfi92: Agreed. Actions presently default to `[]` when there is no actions.

schemas/list-provisioners-response.yml
```diff
@@ -41,6 +41,60 @@ properties:
...
+            description: |
+              Actions are defined on the provisioner. When an action is triggered,
+              it will use the `url` and `context` property to construct the URL and
+              perform a `POST` request. Note that the request will be signed with the
+              user's Taskcluster credentials.
+            properties:
+              label:
+                title:  "Label"
+                description: |
+                  Short names for things like logging/error messages.
```
djmitche: Is this unique?  Is this the one I might match on if I want to tie a specifi action to a button button in Tools?  For task-graph actions we have [`name` for this purpose](https://docs.taskcluster.net/manual/using/actions/spec#metadata)  and some [guidelines for UIs](https://docs.taskcluster.net/manual/using/actions/ui#specialized-triggering).

schemas/provisioner-response.yml
```diff
@@ -39,10 +39,65 @@ properties:
       of when the provisioner was last seen active.
     type:         string
     format:       date-time
-additionalProperties: true
+  actions:
```
djmitche: Could this be factored out into a $constant?

schemas/list-provisioners-response.yml
helfi92: Yes it is unique. I  agree that we should use `name` instead of `label` to respect the UI guidelines.

helfi92:
@djmitche Are you okay if I add documentation to `/docs` in a separate pull-request? Also, where in `/docs`, would it be more appropriate to add documentation about actions.

djmitche:
That should be fine.  All docs/*.md are rendered in the docs site, so just put it in a file with a reasonable name (`actions.md` comes to mind!) and give it some frontmatter similar to https://raw.githubusercontent.com/taskcluster/taskcluster-queue/master/docs/task-life-cycle.md

PR #213, opened 2017-10-17
Bug 1409401 - Implement actions on the Worker Type

helfi92:


helfi92:
I had a meeting with @jonasfj  today and we agreed to let the provisioner be the only entity to declare actions. To that end, actions are to be removed from `declareWorkerType()` and `declareWorker()`.

As a consequence, when a call to `listWorkerTypes()` and `listWorkers()` is made, the two methods will need an extra call to the Provisioner table to extract some of the actions to display. The actions to display will be limited to the same context. In other words, `listWorkerTypes` and `listWorkers` will only fetch the actions with `context=worker-type` and `context=worker` respectively. For this to be done, I will need to wait until https://github.com/taskcluster/taskcluster-queue/pull/212 is merged since the provisioner doesn't have the notion of actions yet.

Meeting notes: https://public.etherpad-mozilla.org/p/jonasfj-hassan-talks-action
/cc @djmitche

helfi92:
Will open a new pull-request.

PR #214, opened 2017-10-19
Fix a typo in schema

jhford:
This is important because it's making implementing taskcluster-lib-artifact-go harder


jhford:
Please merge if you're OK with this fix.  We should consider doing schema validation against our schemas, since json schema publishes a schema.

petemoore:
> Please merge if you're OK with this fix. We should consider doing schema validation against our schemas, since json schema publishes a schema.

Are we definitely not validating? Sadly I think this would passes schema validation with `requiredProperties` too - I think the json schema for json schema doesn't disallow additional properties inside a json schema.

PR #215, opened 2017-10-19
Add actions as top level property for worker-types and workers

helfi92:


helfi92:
Note that actions are only defined on a provisioner but the response value of the worker-type and worker endpoints will return actions of their corresponding context level.

For a provisioner, you get actions of context `provisioner`, `worker-type`, and `worker`. 
For a worker-type, you get actions of context `worker-type`.
For a worker, you get actions of context `worker`.

schemas/list-workers-response.yml
```diff
@@ -60,6 +60,44 @@ properties:
...
       additionalProperties: false
+  actions:
+    type:         array
+    items:
+      type:       object
+      title:      Actions
+      description: {$const: action-description}
+      properties:
+          name:
+                    title:  "Name"
```
jonasfj: 2 spaces indentation...

helfi92:
Will merge tomorrow morning.

PR #219, opened 2017-10-31
Fix declareWorker scope

jonasfj:
I assume this was a typo in the initial implementation...

@helfi92, if this is used by any workers we might have re-issue the scope...
But I don't see anything here: https://tools.taskcluster.net/auth/scopes
so we can probably just deploy...

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -2463,7 +2463,7 @@ api.declare({
   stability:  API.stability.experimental,
   scopes:     [
     [
-      'queue:declare-worker:<provisionerId>/<workerType>/<workerGroup><workerId>#<property>',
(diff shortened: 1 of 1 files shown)


helfi92:
@jonasfj I believe this was done on purpose. You need both a workerGroup and a workerId to identify a worker so it didn't make sense to separate them.

jonasfj:
They are not fixed length...

Consider two workers:
```
workerA:
workerGroup: a
workerId: bb

workerB:
workerGroup: ab
workerId: b
```
These are not the same workers... yet, they have the same scope in this pattern.

helfi92:
Agreed! separators would be helpful in this scenario.

helfi92:
@djmitche I remember we had a discussion regarding that. I tend to agree with what @jonasfj says. If you think otherwise, let us know :)

PR #221, opened 2017-11-03
Disable cloud-mirror for blobs for the time being.

jhford:
Basically, I need to double check a few things in cloud-mirror before we start
sending artifacts to it.  Also adding some headers to the 303s that we're
sending out to include details about the artifact.  This let's us get the
information without actually following a redirect and let's us do sanity checks
in clients that the artifact we originally asked for matches the one we end up
getting, even if there are many redirects in between.

### DIFF
--- a/src/artifacts.js
+++ b/src/artifacts.js
@@ -502,12 +502,26 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
     return res.reportError('ResourceNotFound', 'Artifact not found', {});
   }
 
+  // Some downloading utilities need to know the artifact's storage type to be
+  // able to handle their downloads most correctly.  We're going to set this
+  // field on all artifact responses so that the downloading utilities can use
+  // slightly different logic for each artifact type
+  res.set('x-taskcluster-artifact-type', artifact.storageType);
+
   if (artifact.storageType === 'blob') {
     // Most of the time, the same base options are used.
     let getOpts = {
       bucket: artifact.details.bucket,
       key: artifact.details.key,
     };
+
+    res.set('x-taskcluster-content-sha256', artifact.details.contentSha256);
+    res.set('x-taskcluster-content-length', artifact.details.contentLength);
+    res.set('x-taskcluster-transfer-sha256', artifact.details.transferSha256);
+    res.set('x-taskcluster-transfer-length', artifact.details.transferLength);
+    res.set('x-taskcluster-transfer-length', artifact.details.transferLength);
+    res.set('content-encoding', artifact.details.contentEncoding || 'identity');
+    res.set('content-type', artifact.details.contentType);
 
     // TODO: We should consider doing a HEAD on all resources and verifying that
     // the ETag they have matches the one that we received when creating the artifact.


src/artifacts.js
```diff
@@ -519,39 +529,7 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
-          region,
-          encodeURIComponent(canonicalUrl),
-        ].join('/');
-        return res.redirect(303, urllib.format({
-          protocol: 'https:',
-          host: this.cloudMirrorHost,
-          pathname: cloudMirrorPath,
-        }));
-      }
+      return res.redirect(303, await this.s3Controller.generateGetUrl(getOpts));
```
jonasfj: r+ on this change.

src/artifacts.js
```diff
@@ -509,6 +509,16 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
       key: artifact.details.key,
     };
 
+    res.set('x-amz-meta-runid', runId);
+    res.set('x-amz-meta-taskid', taskId);
+    res.set('x-amz-meta-name', name);
+    res.set('x-amz-meta-content-sha256', artifact.details.contentSha256);
+    res.set('x-amz-meta-content-length', artifact.details.contentLength);
+    res.set('x-amz-meta-transfer-sha256', artifact.details.transferSha256);
+    res.set('x-amz-meta-transfer-length', artifact.details.transferLength);
```
jonasfj: Can we have please have a separate discussion about this...
I'm not sure it's a good idea to do the `x-amz-meta-` headers in our code...

I'm even less sure that it is wise to store things like `transfer-sha256`.
And I see `runid` and `taskid` as useless here...

If you want to add a `x-taskcluster-redirect-content-sha256` (or similar name) I would fully support that. But this seems like it's lacking separation of concerns, coupling too many things too tightly. Are we even storing these headers; I really don't want to do that...

This seems like we coupling the implemention of server and client too tightly, we won't be able to migrated from S3, should we ever need to use a different blob store.

package.json
```diff
@@ -27,7 +27,7 @@
     "netmask": "^1.0.5",
     "promise": "^7.0.4",
     "pulse-publisher": "^1.3.1",
-    "remotely-signed-s3": "^2.0.0",
+    "remotely-signed-s3": "^3.0.0",
```
jonasfj: This was r+'ed in another PR

src/artifacts.js
```diff
@@ -509,6 +509,16 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
       key: artifact.details.key,
     };
 
+    res.set('x-amz-meta-runid', runId);
+    res.set('x-amz-meta-taskid', taskId);
+    res.set('x-amz-meta-name', name);
```
jonasfj: I see where these are set... I don't think they are useful, and I don't think that client should use them any way..

They might nice to have if scanning the S3 bucket.

src/artifacts.js
```diff
@@ -509,6 +509,16 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+    res.set('x-amz-meta-content-sha256', artifact.details.contentSha256);
```
jonasfj: I can't see anywhere this header is set?

src/artifacts.js
```diff
@@ -509,6 +509,16 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+    res.set('x-amz-meta-content-length', artifact.details.contentLength);
+    res.set('x-amz-meta-transfer-sha256', artifact.details.transferSha256);
```
jonasfj: transfer is unimportant, this confuses me a lot...

I assume the only reason we store this in `artifact.details.transferSha256` is to support idempotency in `createArtifact` calls.

jonasfj:
@jhford,
I think it's worthwhile to have `taskcluster-queue` return a header like `x-taskcluster-location-content-sha256` in the redirect.

And I think it's reasonable to document that client SHOULD validate that `x-taskcluster-location-content-sha256` matches the hash of the file they are redirected to.
We can even agree that `x-taskcluster-location-content-sha256` MUST be present when redirecting for `'blob'` artifacts.

I also won't object to a recommendation that a response MAY contain `x-amz-meta-content-sha256` which the client MAY validate. But if we just have the `x-taskcluster-location-content-sha256` header, then that is sufficient and we don't risk make things more complicated than they have to be.

---

I'm scared when I see so many things exposed. I don't want anyone to require the presence of `x-amz-` headers. These won't be supported in other cloud providers. And when `x-taskcluster-location-content-sha256` is sufficient, then let's absolutely not expose more information than necessary, it's a recipe for someone to depend on something that is NOT part of our contract.

jonasfj:
@jhford,
just to be clear, I absolutely LOVE the idea that if clients see a `<...>-content-sha256` from `taskcluster-queue` before they are redirected, then they SHOULD validate it against what they download from the URL they are redirected to.

I think that provides great security, in that you would have to compromise not just S3, but also `taskcluster-queue`. It locks the truth about artifacts in the queue, and I see this as extremely powerful.

I'm only concerned that we're exposing too much metadata that we can't make available across cloud providers... and this leads to clients that expect a contract we're not fully committed to.

package.json
jhford: yep, but not merged in yet, so it showed up here as well

src/artifacts.js
jhford: runid, taskid and name being stored on the actual backing artifact is *really* nice because it allows us to easily rebuild the data about the artifacts stored, and because it is ensured to be correct by remotely-signed-s3 request signing, we can use it to repopulate the datastore, should we lose it.

As well, it allows someone who gets an opaque url to programatically know the name, taskid and runid of the resulting artifact without having needed to speak to the queue.  This would be useful doing debugging around integration of artifacts into the build system possibly.

I don't care so much about the name, but I will say that the overhead to change x-amz-meta to x-azure-meta on an informational header on a redirect would be roughly 0.00000001% of the effort of supporting a second artifact backend.  If we do rename it, let's go with "X-Taskcluster-%s" and drop the unnecessary '-redirect' token.

This is not tight coupling.  These are informational headers, not something stored.  It's totally possible to use a field called 'x-amz-meta-content-sh256' in an azure system.  It's a name, and we can add/change it as needed.

src/artifacts.js
jhford: They are for the purposes of scanning the bucket and also in case we start getting artifacts and can't figure out where they're from.  Having these as headers makes that simple, rather than having to possibly hash the transfer-sh256 and scan the whole artifacts table looking for an artifact that has the transfer-256 that matches, we can simply look at the header.  Since the creation of these headers are signed by remotely-signed-s3, we can trust them to be correct.

src/artifacts.js
jhford: https://github.com/taskcluster/taskcluster-queue/blob/f53a106b7768a8bcd9b01e7664ad3a00575bc8a9/src/artifacts.js#L204

src/artifacts.js
jhford: No, the transfer is actually the only important one in this case.  Cloud-mirror, or any other redirecter is only dealing ever in transfer-bytes.  It'd be silly to have cloud-mirror gunzip the artifact so that it could immediately re-gzip the file.  It is also critical to ensure that the exact bytes which were download are uploaded into the other region.

This has nothing to do with idempotency on createArtifact.

jhford:
This is not about totally rewriting the flow for verifying artifacts, this is merely about adding an extra validation step.  We have clients already written, and if desired, future clients could use these headers.

src/artifacts.js
jonasfj: > we can use it to repopulate the datastore, should we lose it.

That's a solid argument... I'm also okay with store the metadata.

But why do we want to expose it during the redirect? That seems overkill.
Except for exposing the hash of what we're redirecting to, that seems nice.

Okay, I'll try to not mind informational headers :) hehe

src/artifacts.js
jonasfj: That doesn't explain why they are part of the redirect request.


Having them in S3 does no harm. Here it just seems weird to have them. I particularly don't like having things like `x-amz-...` returned from our APIs.

src/artifacts.js
jonasfj: that sets `x-amz-content-sha256`

src/artifacts.js
jonasfj: > Cloud-mirror, or any other redirecter is only dealing ever in transfer-bytes

Fair argument, that makes a lot of sense to me...

Can we please rename this `x-taskcluster-location-transfer-sha256`, as it's a header that indicates the `transfer-sha256` of the `location` header, and this is a taskcluster specific concept.

jonasfj:
> I'm really unclear how this could lead to a contract we can't commit to. Would you care to expand on that?

The queue returning headers on the form: `x-amz-..` is S3 specific, why would it do that...

src/artifacts.js
jhford: so this is one of those things that I want for easier debugging.  I know it's wasted bytes and I know it's not needed, but things like this can make debugging a lot easier.

jhford: sounds like we're in agreement.  I'll do some renames and submit a new PR

src/artifacts.js
jhford: it's just easier to have the same name for the comparisons if we use the same name.

src/artifacts.js
jhford: nope, x-amz-content-sha256 is the sha256 of the request for the S3 request but is not persisted as an object header.  It is actually closer in relation to x-amz-meta-transfer-sha256 than x-amz-meta-content-sha256.  In single part uploads, it does match exactly x-amz-meta-transfer-sha256 but in multipart it matches per request the sha256 of the content-encoded parts.

jhford:
Actually, let's do this as part of https://github.com/taskcluster/taskcluster-queue/pull/233

PR #222, opened 2017-11-03
Bug 1414364 - Switch the disabled property of a worker to quarantineUntil

helfi92:


src/api.js
```diff
@@ -2360,8 +2360,8 @@ api.declare({
   description: [
     'Get a list of all active workers of a workerType.',
     '',
-    '`listWorkers` allows a response to be filtered by the `disabled` property.',
-    'To filter the query, you should call the end-point with `disabled` as a query-string option.',
+    '`listWorkers` allows a response to be filtered by the `quarantineUntil` property.',
+    'To filter the query, you should call the end-point with `quarantineUntil` as a query-string option.',
```
djmitche: What does this return?  Those quarantined before that date? after?

src/api.js
```diff
@@ -2379,12 +2379,9 @@ api.declare({
     provisionerId,
     workerType,
     expires: Entity.op.greaterThan(new Date()),
+    quarantineUntil: Entity.op.greaterThan(new Date(quarantineUntil || -8640000000000000)),
```
djmitche: probably better to not include this condition when it's not specified - it might allow Azure to return results more efficiently.

src/api.js
```diff
@@ -2455,7 +2452,52 @@ api.declare({
...
+  method: 'put',
+  route:  '/provisioners/:provisionerId/worker-types/:workerType/workers/:workerGroup/:workerId',
+  name:   'quarantineWorker',
+  stability: API.stability.experimental,
+  scopes: [
+    [
+      'queue:quarantine-worker:<provisionerId>/<workerType>/<workerGroup>/<workerId>',
+    ],
+  ],
+  deferAuth: true,
```
djmitche: I don't think this is necessary here -- all of the scopes' parameters are already available.

src/api.js
```diff
@@ -2399,7 +2396,7 @@ api.declare({
         workerId: worker.workerId,
         firstClaim: worker.firstClaim.toJSON(),
         latestTask: worker.recentTasks.pop(),
-        disabled: worker.disabled,
+        quarantineUntil: worker.quarantineUntil.toJSON(),
```
djmitche: we should probably omit this when the date is in the past, so that users can simply check for existence when wondering if a worker is currently quarantined.

djmitche: Also, could this use `worker.json()`?

src/data.js
```diff
@@ -960,7 +984,7 @@ Worker.prototype.json = function() {
     provisionerId:    this.provisionerId,
     workerId:         this.workerId,
     workerGroup:      this.workerGroup,
-    disabled:         this.disabled,
+    quarantineUntil:  this.quarantineUntil.toJSON(),
```
djmitche: as above, let's omit this if quarantineUntil < now

src/data.js
```diff
@@ -933,6 +933,30 @@ let Worker = Entity.configure({
...
+     * See JSON schema for documentation.
+     */
+    recentTasks:      Entity.types.JSON,
+    quarantineUntil:  Entity.types.Date,
+    // the time at which this worker should no longer be displayed
+    expires:          Entity.types.Date,
+    firstClaim:       Entity.types.Date,
+  },
+  migrate(item) {
+    item.quarantineUntil = new Date();
```
djmitche: I'm a little worried about clock skew here.  Could we default these to 1970 or something like that?

helfi92: Most definitely. `Date(0)` will set it to `1970-01-01T00:00:00.000Z`.

src/api.js
helfi92: after

src/api.js
djmitche: Did this get missed?

schemas/quarantine-worker-request.yml
```diff
@@ -0,0 +1,14 @@
+$schema:  http://json-schema.org/draft-04/schema#
+title:          "Quarantine Worker Request"
+description: |
+  Request to update a worker's quarantineUntil property.
+type:           object
+properties:
+  quarantineUntil:
+    title:      "Worker Quarantine"
+    description: |
```
jonasfj:
```
type: string
format: date-time
```

schemas/list-workers-response.yml
```diff
@@ -26,12 +26,13 @@ properties:
           minLength:    {$const: identifier-min-length}
           maxLength:    {$const: identifier-max-length}
           pattern:      {$const: identifier-pattern}
-        disabled:
-          title:        "Worker Disabled"
+        quarantineUntil:
+          title:        "Worker Quarantine"
```
jonasfj:
```
type: string
format: date-time
```

src/api.js
```diff
@@ -2351,7 +2351,7 @@ api.declare({
   query: {
     continuationToken: /./,
     limit: /^[0-9]+$/,
-    disabled: /^(true|false)$/,
+    quarantineUntil: /\d{4}-[01]\d-[0-3]\dT[0-2]\d:[0-5]\d:[0-5]\d\.\d+([+-][0-2]\d:[0-5]\d|Z)/,
```
jonasfj: Might it not be safer to say `quarantined: true|false`

Timestamps are complicated... and this is unlikely to match timestamps produced in python and other languages..
I'm pretty sure we use "ISO 8601" with "Z" for UTC time... but it gets complicated when some timestamps have microseconds and what not... Are you sure we covered all that here?

Note: I think `quarantined: true|false` is complicated enough. We can always add more advanced filters later.

src/api.js
```diff
@@ -2394,13 +2395,17 @@ api.declare({
...
-        disabled: worker.disabled,
-      };
+      return Object.assign(
+        {
+          workerGroup: worker.workerGroup,
+          workerId: worker.workerId,
+          firstClaim: worker.firstClaim.toJSON(),
+          latestTask: worker.recentTasks.pop(),
+        },
+        worker.quarantineUntil.getTime() > new Date().getTime() ?
```
jonasfj: Are you modeling `quarantineUntil` as an optional property?

Is that wise? compared to either setting it `null` or perhaps even better setting to `'1970-01-01T00:00:00.000Z'`.
In fact if it was quarantined in the past is there anything wrong with having that datetime in there?

djmitche: My reasoning is that we are treating a quarantineUntil date in the past as meaning "not quarantined".  I don't think we should require the caller to do that datetime comparison since the most common reason to look at this field will be to answer "is it quarantined right now".  I wouldn't mind setting it to null instead of omitting the field.

jonasfj: hmm, fair enough... I buy this argument.

datetime comparisons are hard :)

src/api.js
helfi92: @jonasfj This will cover `ISO 8601` (e.g., `new Date().toISOString()`). Do you still want me to change `quarantineUntil` to `quarantined`? Also, do you only want that changed for `listWorkers()` only or for all endpoints returning `qurantineUntil`?

djmitche: I think it's the parameter in particular Jonas is concerned about.  We don't typically offer very sophisticated filtering arguments on our listXxx methods, so this seems odd.  And it will reject ISO 8601 timestamps that fall exactly on the second, as those do not have a `.` in them.

helfi92: My last commit changes quarantineUntil to quarantined :)

helfi92:
Just spoke with @gregarndt and we agreed that we should probably wait until next tuesday (launch day) before merging this because of change freeze.

PR #223, opened 2017-11-08
Modify action explanations to reflect path parameters

helfi92:
The best option here is to switch actions back to allowing substitution in the URL path. This pull-request modifies the action explanations to talk about path parameters.

### DIFF
--- a/schemas/constants.yml
+++ b/schemas/constants.yml
@@ -132,25 +132,33 @@ dependency-relation:
 
 # Description of actions. They are defined on the provisioner
 action-description: |
-  Actions are defined on the provisioner. When an action is triggered,
-  it will use the `url` and `context` property to construct the URL and
-  perform a `POST` request. Note that the request will be signed with the
-  user's Taskcluster credentials.
+  Actions are defined on the provisioner and they are performed via a `POST` request using the `url`.
+  Depending on the `context`, the `url` will be able to fill path parameters. In addition, `context` is
+  used by the front-end to know where to display the action. For example, `context=worker`
+  will display the action on the worker explorer.
+
+  _Note: The request will be signed with the user's Taskcluster credentials._
 
 # Description of the action context
 action-context-description: |
   Actions have a "context" that is one of provisioner, worker-type,
-  or worker, indicating which it applies to. `context` is used to construct
-  the query string of the `POST` request.
-  If `context='worker'`, the query string will be
-  `?provisionerId=${PROVISIONER_ID}&workerType=${WORKER_TYPE}&workerGroup=${WORKER_GROUP}&workerId=${WORKER_ID}`.
-  If `context='worker-type'`, the query string will be
-  `?provisionerId=${PROVISIONER_ID}&workerType=${WORKER_TYPE}`.
-  If `context='provisioner'`, the query string will be
-  `?provisionerId=${PROVISIONER_ID}`.
+  or worker, indicating which it applies to. `context` is used by the front-end to know where to display the action.
+
+  | `context`   | Path parameters                                      | Page displayed        |
+  |-------------|------------------------------------------------------|-----------------------|
+  | provisioner | :provisionerId                                       | Worker explorer       |
+  | worker-type | :provisionerId, :workerType                          | Workers explorer      |
+  | worker      | :provisionerId, :workerType, :workerGroup, :workerId | Worker types explorer |
 
 # Description of the url property in an action.
 action-url-description: |
-  When an action is triggered, the `url`
-  and `context` property are used to make the `POST` request.
-  The request needs to be signed with the user's Taskcluster credentials.
+  When an action is triggered, a `POST` request is made using the `url`.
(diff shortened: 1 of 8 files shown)


helfi92:
I will modify the docs in a different pull-request.

schemas/constants.yml
```diff
@@ -132,25 +132,33 @@ dependency-relation:
 
 # Description of actions. They are defined on the provisioner
 action-description: |
-  Actions are defined on the provisioner. When an action is triggered,
-  it will use the `url` and `context` property to construct the URL and
-  perform a `POST` request. Note that the request will be signed with the
-  user's Taskcluster credentials.
+  Actions are defined on the provisioner and they are performed via a `POST` request using the `url`.
+  Depending on the `context`, the `url` will be able to fill path parameters. In addition, `context` is
```
djmitche: ..will be filled with parameters

schemas/constants.yml
```diff
@@ -132,25 +132,33 @@ dependency-relation:
...
-  `?provisionerId=${PROVISIONER_ID}&workerType=${WORKER_TYPE}&workerGroup=${WORKER_GROUP}&workerId=${WORKER_ID}`.
-  If `context='worker-type'`, the query string will be
-  `?provisionerId=${PROVISIONER_ID}&workerType=${WORKER_TYPE}`.
-  If `context='provisioner'`, the query string will be
-  `?provisionerId=${PROVISIONER_ID}`.
+  or worker, indicating which it applies to. `context` is used by the front-end to know where to display the action.
+
+  | `context`   | Path parameters                                      | Page displayed        |
+  |-------------|------------------------------------------------------|-----------------------|
+  | provisioner | :provisionerId                                       | Worker explorer       |
```
djmitche: Provisioner Explorer

schemas/constants.yml
```diff
@@ -132,25 +132,33 @@ dependency-relation:
...
+  | worker-type | :provisionerId, :workerType                          | Workers explorer      |
```
djmitche: Worker-Type Explorer

schemas/constants.yml
```diff
@@ -132,25 +132,33 @@ dependency-relation:
...
+  | worker      | :provisionerId, :workerType, :workerGroup, :workerId | Worker types explorer |
```
djmitche: Worker Explorer

schemas/constants.yml
djmitche: Also, you can probably omit the path parameters here, as they are included in action-url-description.

schemas/constants.yml
```diff
@@ -132,25 +132,33 @@ dependency-relation:
...
 
 # Description of the url property in an action.
 action-url-description: |
-  When an action is triggered, the `url`
-  and `context` property are used to make the `POST` request.
-  The request needs to be signed with the user's Taskcluster credentials.
+  When an action is triggered, a `POST` request is made using the `url`.
+  Depending on the `context`, the `url` will be able to fill the following path parameters:
```
djmitche: ..the following parameters will be substituted in the `url`:

test/workerinfo_test.js
```diff
@@ -117,13 +117,13 @@ suite('provisioners and worker-types', () => {
...
         context: 'provisioner',
-        url: 'https://hardware-provisioner.mozilla-releng.net/v1/power-cycle',
+        url: 'https://hardware-provisioner.mozilla-releng.net/v1/power-cycle/:provisionerId',
         description: 'Remove provisioner prov-B',
       }, {
         name: 'kill',
         title: 'Kill Worker Type',
         context: 'worker-type',
-        url: 'https://hardware-provisioner.mozilla-releng.net/v1/power-cycle',
+        url: 'https://hardware-provisioner.mozilla-releng.net/v1/power-cycle/:provisionerId/:workerType',
```
jonasfj: Against please consider `<provisionerId>` instead of `:provisionerId` as the parameter format... this is more in line with what we do in `references.json`..

helfi92:
@jonasfj @djmitche  I added a method property to the action definition. I also removed `format: uri` from the action `url` since we decided to use something like `<provisionerId>` instead of `:provisionerId`.

schemas/list-workers-response.yml
```diff
@@ -85,7 +85,12 @@ properties:
           title:  "URL"
           description: {$const: action-url-description}
           type:    string
-          format:  uri
+        method:
+          title:  "Method"
+          description: |
+            Method to indicate the desired action to be performed for a given resource.
+          type:   string
+          enum:   ["GET", "POST", "PUT", "DELETE", "PATCH"]
```
djmitche: I don't think GET should ever perform an action.

helfi92: This must have slipped in by accident. Fixing that right now!

helfi92:
@jonasfj I will merge now but if there's anything you see that needs to be added/removed, let me know :)

PR #224, opened 2017-11-08
Remove actions query parameters from docs

helfi92:


helfi92:
@djmitche I'm trying to see if we can merge this today if possible. We don't want people to start planning on the older design of queries.

docs/actions.md
```diff
@@ -18,42 +18,47 @@ An action is comprised with the following properties:
...
 
 ## How to trigger an action
-To trigger an action, make a `POST` request to a URL constructed from the url field with query
-arguments added based on the `context`.
+To trigger an action, use the action's `url` property to make a `POST` request to it.
+Depending on the action's `context`, substitute the following parameters in the `url`:
+
+| `context`   | Path parameters                                      |
+|-------------|------------------------------------------------------|
+| provisioner | :provisionerId                                       |
```
jonasfj: Can we use `<provisionerId>`  like we do in `entries[i].route` in `reference.json` files, see docs:
https://docs.taskcluster.net/manual/design/apis/reference-format

djmitche:
+1 with jonas's suggestion

helfi92:
Done! Will merge later today.

PR #227, opened 2017-11-09
Hotfix remove uri format from schema since we're using <parameter>

helfi92:


schemas/worker-response.yml
```diff
@@ -114,6 +113,7 @@ properties:
         - title
         - context
         - url
+        - method
```
petemoore: @helfi92 There is no property defined called `method` :/

PR #230, opened 2017-11-16
Expand Queue Actions documentation

djmitche:
This
 * defines the provisioner / worker-type / worker hierarchy a little
   better
 * distinguishes queue actions from task actions
 * removes specific reference to the tools UI views
 * gives more detail on how calls can occur from automation, including
   use of the `name` property

It presupposes some planned changes to the queue APIs; in particular,
that actions be returned from the getXxx methods and not from the
listXxx methods.

### DIFF
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -1,16 +1,30 @@
 ---
 title: Actions
-description: How actions are defined on the provisioner.
+description: Queue Actions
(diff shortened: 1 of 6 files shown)


djmitche:
Good point, thanks

djmitche:
updated

PR #231, opened 2017-11-16
Remove actions when listing workerTypes and workers

helfi92:
Actions are not needed when listing workers or worker-types. When a worker-type requests a list of workers for example, the worker-type will then need to make a request to `getWorkerType()` to get the list of actions with  `context=worker-type`.

### DIFF
--- a/schemas/list-workers-response.yml
+++ b/schemas/list-workers-response.yml
@@ -63,50 +63,6 @@ properties:
               minimum:          {$const: min-run-id}
               maximum:          {$const: max-run-id}
       additionalProperties: false
-  actions:
-    type:         array
-    items:
-      type:       object
-      title:      Actions
-      description: {$const: action-description}
-      properties:
-        name:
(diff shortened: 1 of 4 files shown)


helfi92:
@djmitche It can but I have https://github.com/taskcluster/taskcluster-tools/pull/382 set up to fix this. Would you recommend me to merge these two pull-requests during the weekend instead?

djmitche:
If the tools patch lands first things should just keep working, right?  Or, yes, landing both during a period of low use sounds fine.

helfi92:
@djmitche Sorry I thought this was https://github.com/taskcluster/taskcluster-queue/pull/222. This shouldn't break the existing tools implementation. It's not fetching the actions from `listWorkers` nor `listWorkerTypes`.

djmitche:
Cool!  Then ship this when you're ready and comfortable with the change-freeze implications..

PR #232, opened 2017-11-22
Make property optional rather than returning null value

petemoore:
Hi @helfi92,

Although it is entirely valid to specify `type` as the array `["string", "null"]` I think the advantage of making it an optional property is that if there is no value, `quarantineUntil` does not needed to be included, rather than `"quarantineUntil": null`.

I haven't tested this change, but I'm hoping the CI will take care of this, or you'll be able to let me know if I need to do any additional testing.

Note: the motivation for making this change is because it broke the code generation in taskcluster-client-go. Until now, we never had a type that permitted multiple values. I could resolve this by enhancing taskcluster-client-go too (which I will probably do) but in discovering this, I saw it was inconsistent with the way we manage optional fields in the rest of the project, so I thought it would make sense to align our schemas on how they handle optional values, to give a consistent treatment across our APIs/data structures.

Let me know what you think, or if you want to discuss it!

Thanks!

### DIFF
--- a/schemas/list-workers-response.yml
+++ b/schemas/list-workers-response.yml
@@ -33,7 +33,7 @@ properties:
             Once the quarantineUntil time has elapsed, the worker resumes accepting jobs.
             Note that a quarantine can be lifted by setting `quarantineUntil` to the present time (or
             somewhere in the past).
-          type:         ["string", "null"]
+          type:         string
           format:       date-time
         firstClaim:
           title:        "First task claimed"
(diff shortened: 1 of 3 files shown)


petemoore:
@djmitche back to you :-)

petemoore:
And thanks John for the commits!

djmitche:
I'm backing this out as it [failed tests](https://travis-ci.org/taskcluster/taskcluster-queue/builds/307993126)

petemoore:
> I'm backing this out as it [failed tests](https://travis-ci.org/taskcluster/taskcluster-queue/builds/307993126)

This failure was unrelated to the change and at first glance looks likely to be a race condition in the test itself with a 2 second delay that might be too aggressive. Retriggering travis job resulted in a successful test run. I'll backout the backout tomorrow morning. ;-)

PR #233, opened 2017-11-22
bug 1419577 -- URL Encode the prefix portion of S3 urls

jhford:
It turns out that regardless of what we put into the prefix, S3 will URL Encode
some characters of the prefix (e.g. + will always be encoded to %2B).  This
results in broken URLs.  This approach is the simple approach of just encoding
the entire S3 Object Key.  I've verified that with a problem URL (e.g. has + in
it) that this encoding results in valid URLS which get 200 from S3 instead of
403s.  I've also verified that Cloudfront also supports this encoding.

Another approach for prettier URLs would be to revert the %2F character back to
a / character since S3 and CloudFront already deal with those characters quite
well.  This would require prefix = encodeURIComponent(prefix).replace(/%2f/gi,
"/");

This also includes patches for cleaning up handling of magic headers as well as
adding a couple of headers to the artifact responses


jhford:
@walac fyi, 325ece6 is needed in order to support non-blob artifacts in taskcluster-lib-artifact

src/artifacts.js
```diff
@@ -502,13 +502,27 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
     let getOpts = {
       bucket: artifact.details.bucket,
       key: artifact.details.key,
     };
 
+    res.set('x-taskcluster-content-sha256', artifact.details.contentSha256);
+    res.set('x-taskcluster-content-length', artifact.details.contentLength);
+    res.set('x-taskcluster-transfer-sha256', artifact.details.transferSha256);
+    res.set('x-taskcluster-transfer-length', artifact.details.transferLength);
+    res.set('x-taskcluster-transfer-length', artifact.details.transferLength);
```
jonasfj: Can we maybe settle for having the length being present only once? :)

src/artifacts.js
```diff
@@ -502,13 +502,27 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+    res.set('content-encoding', artifact.details.contentEncoding || 'identity');
+    res.set('content-type', artifact.details.contentType);
```
jonasfj: So these could confuse a client... let's make them `x-taskcluster-location-content-type`

src/artifacts.js
```diff
@@ -502,13 +502,27 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+  res.set('x-taskcluster-artifact-type', artifact.storageType);
+
   if (artifact.storageType === 'blob') {
     // Most of the time, the same base options are used.
     let getOpts = {
       bucket: artifact.details.bucket,
       key: artifact.details.key,
     };
 
+    res.set('x-taskcluster-content-sha256', artifact.details.contentSha256);
```
jonasfj: These are headers for the thing we are redirecting to, hence, I suggest they are on the form:
`x-taskcluster-location-`

like `x-taskcluster-location-content-sha256` is the `content-sha256` of what the `location` header points to.

src/artifacts.js
```diff
@@ -502,13 +502,27 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
     return res.reportError('ResourceNotFound', 'Artifact not found', {});
   }
 
+  // Some downloading utilities need to know the artifact's storage type to be
+  // able to handle their downloads most correctly.  We're going to set this
+  // field on all artifact responses so that the downloading utilities can use
+  // slightly different logic for each artifact type
+  res.set('x-taskcluster-artifact-type', artifact.storageType);
```
jonasfj: I would like to avoid this, it's rather implementation specific, and we have many clients.

I think we should document headers like `x-taskcluster-location-content-sha256`. For example we could say:

---
A taskcluster service MAY supply a `x-taskcluster-location-content-sha256` header when redirecting clients. If present clients SHOULD validate the content of the what is being downloaded from the `location` header against hexadecimal encoded sha256 from this header.

---

Or something like that... This way if we change the implementation, we can keep using the header... And if in the future we only use blob artifacts, we can change the documentation to say the header MUST be present, and SHOULD be validated.

Similarly, we can document other headers and how they should be handled. If we decide we need `sha3` in some future, it would be awesome that we don't have to invent a new artifact type. We just gracefully add it and later make it required. Or if we invent a new artifact type, then we can keep adding the `sha256` headers for compatibility with older clients; and they won't stop using it just because the implementation changed.

----
Note: we should probably write a page on the docs site where we explain these headers, what they are for and use words like MAY and SHOULD. Probably we could have a doc that says HTTP clients that use taskcluster APIs MUST follow redirects, and MUST NOT resend the `authentication` header when following a redirect. We haven't really documented this yet, we just sort of take it for granted -- pretty sure we agreed on things like this when jlal was around, but that's years ago.. maybe we should write it down?

As long as you keep reminding me that `transfer-sha256` headers are useful for supporting pass-through caches I think they are great. But maybe this is a good thing to write down in such docs as well.. As a side note that this is why it's useful :)

src/bucket.js
```diff
@@ -98,6 +99,7 @@ Bucket.prototype.createSignedGetUrl = function(prefix, options) {
   assert(prefix,                'prefix must be given');
   assert(options,               'options must be given');
   assert(options.expires,       'expires must be given');
+  prefix = encodeURIComponent(prefix);
   return new Promise((accept, reject) => {
     this.s3.getSignedUrl('getObject', {
       Key:          prefix,
```
jonasfj: Won't aws-sdk do the right thing here?

src/bucket.js
```diff
@@ -80,6 +80,7 @@ Bucket.prototype.createPutUrl = function(prefix, options) {
  */
 Bucket.prototype.createGetUrl = function(prefix, forceS3 = false) {
   assert(prefix, 'prefix must be given');
+  prefix = encodeURIComponent(prefix);
```
jonasfj: I'm surprised it's worked this long without this... Really this should be a separate PR... and we should be on hand to monitor what happens when you land this...

Well, at least, a separate PR so that it's easy to back out, and then ping sherifs to look out for weird artifact errors while you deploy. You have my **r+** on this change. It seems like you did the due diligence on testing things.

src/artifacts.js
jhford: heh, yah

src/artifacts.js
jhford: fair point, sure!

src/bucket.js
jhford: it's worked so long because the vast majority of cases where there'd be a URI Encoding substitution don't cause an issue with S3 (e.g. / chars).  It's weird how S3 accepts a partial URI encoding.

A second PR doesn't improve the deployment or backout options really.  This is already its own commit, so the backout would be `git revert 227d7aebd0563ec6fd0ccc1ced5d4f85fc7dbbb6`.

src/artifacts.js
jhford: Specifying the artifact *storage type* is not implementation specific at the level being suggested.  This is an implementation specific feature of the *entire* artifact subsystem, not of any specific type.  Specifying this value is important because certain clients (e.g. `taskcluster-lib-artifact{,-go}`) need to be able to tell the difference between the various artifact types.  When downloading an artifact, this library needs to know whether to do validation or not.  For a blob storage artifact, these headers are not an optional component.  They are 100% mandatory, so we should be validating that when we have a blob artifact, that they are present.  

Because we don't have these hash/size values for the S3 or Azure blob type, nor can we, we need to be able to ignore validation for those two artifact types.  For Error and Redirect artifacts, we would either need to start storing the SHA256 + Size of the response body at creation or when we send it.  How would we handle the validation of a redirect which can change?  We have different storage types, and each storage type might need slight different handling.  We could imply artifact type on response, but why?  We can easily be explicit without any detriment.

The differences between using Sha256 and a Sha-3 algorithm *can* be invisibly adapted within the blob artifact.  Which hashing algorithm used by a given storage type does not have *any* bearing on what artifact storage type it is.  It's either a blob, S3, Azure, Error or Redirect.  The change to Sha3 would represent such a large amount of work that changes to this header would be trivial, even if it were related.

I disagree that taskcluster api clients must follow redirects.  I think a more correct phrasing is that the get-artifact endpoint returns a response value which is in the form of a valid http 303 redirect with metadata exposed as X-Taskcluster prefixed headers.

One specific case that would violate this declaration is around HTTPS verification.  The `taskcluster-lib-artifact*` libraries understand that we should never follow redirects to non-https resources.  If we made it mandatory to follow the redirect and received a redirect to an HTTP resource, we'd still have to follow that redirect even if we didn't want the actual body data.

Another case is that someone is doing checks to see if an artifact on a test machine is valid.  The only way to get that information is to do a non-redirect-following call to get-artifact and find the data there.  If we made this mandatory, they'd have to redownload the whole thing just to get the Sha256 value.

Adding this header has literally zero impact on clients which follow the redirect.  It allows those clients which want to opt into certain behaviour the ability to do so explicitly instead of implicitly.  The value which this header represents is such a core concept to the artifact subsystem that any change it in would likely be a complete rewrite.

What are the specific technical concerns with exposing this information?

src/bucket.js
jhford: no.  I keep saying that aws-sdk is not that spectacular.

jhford:
Hi  Jonas, when you have a chance to read through https://github.com/taskcluster/taskcluster-queue/pull/233#discussion_r152805465, it'd be appreciated.

src/artifacts.js
jonasfj: > For a blob storage artifact, these headers are not an optional component. They are 100% mandatory, so we should be validating that when we have a blob artifact, that they are present.

We are largely in agreement, this is a technicality, @jhford let's talk on vidyo.... we can bike shed some other stuff too... I'll send an invite.

> We could imply artifact type on response, but why?

I wouldn't infer the artifact type; I would phrase it as clients should be unaware of the type. But that they SHOULD validate using headers like `x-taskcluster-location-content-sha256` if present.
So you get graceful degrading...

> Another case is that someone is doing checks to see if an artifact on a test machine is valid. 

Ideally that should use HEAD request.

> What are the specific technical concerns with exposing this information?

Tight coupling.... visibility of concepts that could be hidden. Yeah, this is all minor nits.

jonasfj: okay, let's call it `x-taskcluster-storage-type`

src/artifacts.js
```diff
@@ -550,15 +550,33 @@ var replyWithArtifact = async function(taskId, runId, name, req, res) {
...
+      // CDN (cloudfront) for those requests which require it
       let skipCacheHeader = (req.headers['x-taskcluster-skip-cache'] || '').toLowerCase();
-      if (!region) {
-        debug('artifact from CDN for ip: %s', req.headers['x-forwarded-for']);
-        url = this.publicBucket.createGetUrl(prefix);
-      } else if (skipCacheHeader === 'true' || skipCacheHeader === '1') {
-        // Skip cache and go to cloud-front
-        url = this.publicBucket.createGetUrl(prefix);
-      } else if (this.artifactRegion === region) {
+      let skipCDNHeader = (req.headers['x-taskcluster-skip-cdn'] || '').toLowerCase();
```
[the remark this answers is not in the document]
jonasfj: for debugging only...

jhford:
So we had a discussion about this and decided:

1. we'll expose the `storageType` header
1. we'll add `-location` to the other metadata headers
1. we should document these headers and how the download flow should be done in the GetArtifact endpoint documentation.  This should include which encoding the Sha256 values
1. we'll leave the skipCDN 
1. lib-artifact should do feature detection -- but use the `storageType` to handle special cases where specific storage-types

petemoore:
Just wanted to call out there is a test failure here, before this gets merged. :-)

jonasfj:
@jhford, pete has a point about errors... anyways, I doubt you're deploying today :)

Also please reconsider some of logging... did you change `DEBUG`, or something? we is the upstream library printing a lot of unnecessary log messages..

Afaik we can't have too much garbage logging in production as it uses up our logging quota.. And makes log unusable...

jhford:
@jonasfj please take another look at this.  I know that @edmorley would like to be able to use the skip-cdn header and the URI Encoding is still a bug.

Regarding logging, this is a change only to travis logs, so we're not actually using any production logging overhead.

src/bucket.js
jonasfj: I think we need to only encode the minimum number of characters... So the URLs are still pretty...

Things like `/` that don't need encoding should not be encoded...  I suspect there is code that relies on URLs being pretty.

Because of curl or wget that probably does things like parsing the last slash to determine the file name they are writing to...

src/bucket.js
```diff
@@ -97,6 +98,7 @@ Bucket.prototype.createSignedGetUrl = function(prefix, options) {
   assert(prefix,                'prefix must be given');
   assert(options,               'options must be given');
   assert(options.expires,       'expires must be given');
+  prefix = encodeURIComponent(prefix);
```
jonasfj: Can we please avoid escaping anything that doesn't strictly need to be escaped...

Notably, this will cause `/` to be escaped.. let's try to avoid that.

jhford: I'm very reluctant to not fully URI encode things, since we're specifically trying to fix something where we made incorrect assumptions about how keys are URI encoded within S3.

Since these are not queue urls, there's no real reason to make them look pretty, only risk.  The queue urls would still have `/` in them, just that the redirect `Location` headers won't.  Anyone looking at the redirect `Location` likely understands enough to be able to figure out that they're URI encoded.

jonasfj: @jhford, fair point, but please monitor the trees and make an announcement about the risk with this encoding. I'm not convinced that encoding `/` won't cause problems.

jonasfj: @djmitche, can you help make a call on the risk of encoding `/` in the URLs that the queue redirects to...

it's plausible that I'm being irrational here :)

jhford: I think the risk is low, but the reward is so small that it's still not worth it.  Let's deploy this an revisit if we have a lot of complaints about the urls in the `Location` headers

djmitche: I don't know what the original encoding issue was with S3 -- @jhford says "S3 will URL Encode some characters of the prefix" in the intro to this PR, but I don't know when S3 generates URLs..

It's not surprising that S3 accepts "partial URI encoding" -- because URI encoding is identity for so many characters, it can be a funny encoding to deal with.  S3 just decodes any %xx's that it finds, so I'm sure it would happily decode `%4A` to `J`.  That's not really unusual.  Regarding `/` specifically, the S3 backend doesn't treat the character specially, and it doesn't expect any other path components in the key name, so it's just as happy with `%2F` as with `/`.  To "fully" urlencode something would be to produce a string of only `%xx%xx%xx`, and I don't think we should do that.  The question is what characters to encode -- and that needs to be at least the set of characters which the receiving system (S3, in this case?) would interpret as having another meaning.  Anything over that set is unnecessary but probably won't cause issues with that receiving system.

In other words, I don't think it matters to S3.  And it doesn't matter much to users, since they generally won't see these URLs.  History suggests it probably will matter to @emorley though.  And for my own purposes, it's nice to be able to read the URLs in a logfile, and `/foo` is a lot easier to read than `%2Ffoo`.

djmitche:
This looks like a changed-a-bunch-of-stuff PR that changes a bunch of things for which I don't have the background.  I think @jonasfj flagged me for review to answer the question in the ping, so I'll remove my review request.

jonasfj:
@jhford, okay, then let's ship it :)

You probably right that changing the `Location` header will have little impact.

PR #235, opened 2017-11-28
Make quarantineUntil an optional field, rather than allowing null value

petemoore:
This builds on #232 which did not catch all occurrences.

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -979,7 +979,7 @@ Worker.expire = async function(now) {
 };
 
 Worker.prototype.json = function() {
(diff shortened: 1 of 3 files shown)


src/data.js
```diff
@@ -979,18 +979,19 @@ Worker.expire = async function(now) {
 };
 
 Worker.prototype.json = function() {
-  return {
+  let w = {
```
helfi92: I would rename to `const worker = { ... }` to make the variable name descriptive.

PR #239, opened 2017-12-21
Bug 1249045 - issue named temporary credentials to tasks

petemoore:
I haven't run any tests locally, hoping travis will take care of that for me. :-)

Note, since auth now applies clock drift adjustment, this is also no longer necessary in the queue, so have removed that too.

Note, the temp credential naming schema I used is based on the scheme defined already in [our docs](https://github.com/taskcluster/taskcluster-docs/blob/ce34244bd65917485c2120e308c097c9ff5157b9/src/manual/design/namespaces.md#clients).

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -5,6 +5,7 @@ let _           = require('lodash');
 let API         = require('taskcluster-lib-api');
 let Entity      = require('azure-entities');
 let taskcluster = require('taskcluster-client');
+let taskCreds   = require('./task-creds');
 let Promise     = require('promise');
 
 // Maximum number runs allowed
@@ -1649,28 +1650,15 @@ api.declare({
     );
   }
 
-  // Create temporary credentials for the task
-  let clientId = [
-    'task-client',
+  let credentials = taskCreds(
     taskId,
-    `${runId}`,
-    'on',
+    runId,
     run.workerGroup,
     run.workerId,
-    'until',
-    `${takenUntil.getTime() / 1000}`,
-  ].join('/');
-  let credentials = taskcluster.createTemporaryCredentials({
-    clientId,
-    start:  new Date(),
-    expiry: takenUntil,
-    scopes: [
-      'queue:reclaim-task:' + taskId + '/' + runId,
-      'queue:resolve-task:' + taskId + '/' + runId,
-      'queue:create-artifact:' + taskId + '/' + runId,
-    ].concat(task.scopes),
-    credentials: this.credentials,
-  });
+    takenUntil,
+    task.scopes,
+    this.credentials,
+  );
 
   // Reply to caller
   return res.reply({
--- a/src/api.js
+++ b/src/api.js
@@ -1650,9 +1650,20 @@ api.declare({
   }
 
   // Create temporary credentials for the task
+  let clientId = [
+    'task-client',
+    taskId,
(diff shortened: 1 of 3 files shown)


petemoore:
CC @djmitche

src/api.js
```diff
@@ -1651,14 +1651,17 @@ api.declare({
...
+    expiry: takenUntil,
     scopes: [
       'queue:reclaim-task:' + taskId + '/' + runId,
       'queue:resolve-task:' + taskId + '/' + runId,
       'queue:create-artifact:' + taskId + '/' + runId,
     ].concat(task.scopes),
     credentials: this.credentials,
+    clientId:
+      'task-client/' + '/' + runId + '/on/' + run.workerGroup + '/' +
+      run.workerId + '/until/' + takenUntil.getTime(),
```
jonasfj: I propose `task/${taskId}/${runId}@${workerGroup}/${workerId}`, the takenUntil is nice in theory, but will we ever need it in practice...

petemoore:
Something bizarre going on here - I see in [this comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1424986#c0) that it already says:

```
http.Header{"Authorization":[]string{"Hawk id=\"task-client/J1oFCsBcRASocLn_Y-KFmg/0/on/pmoore-own-machines/pmoore-imac/until/1512979345.375\", mac=\"*********************************\", ts=\"1512979721\", nonce=\"*********************************\", ext=\"****************\""}}
```

which suggests this is already implemented in the Queue. Now I'm confused!

petemoore:
> Something bizarre going on here - I see in [this comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1424986#c0) that it already says:
> 
> ```
> http.Header{"Authorization":[]string{"Hawk id=\"task-client/J1oFCsBcRASocLn_Y-KFmg/0/on/pmoore-own-machines/pmoore-imac/until/1512979345.375\", mac=\"*********************************\", ts=\"1512979721\", nonce=\"*********************************\", ext=\"****************\""}}
> ```
>
> which suggests this is already implemented in the Queue. Now I'm confused!

I've just spotted this comes from `claimTask` and I was updating `reclaimTask`. I'll make them consistent.

src/api.js
```diff
@@ -1650,9 +1650,20 @@ api.declare({
...
+  let clientId = [
+    'task-client',
+    taskId,
+    `${runId}`,
+    'on',
+    run.workerGroup,
+    run.workerId,
+    'until',
+    `${takenUntil.getTime() / 1000}`,
+  ].join('/');
```
djmitche: Let's pull this out to a utility function since it now appears in two places?

petemoore: You're quite right. There might be other common code between `claimTask` and `reclaimTask` that could be shared too - I'll take a look.

djmitche:
Still need to fix up the tests...

petemoore:
> Still need to fix up the tests...

Done! :-)

petemoore:
@djmitche I've no idea if I've done this idomatically (putting it in its own module) - let me know if it would be better done in a different way. This was just the easiest way I saw to make the function visible to both places that use it.

djmitche:
Looks idiomatic to me!

PR #242, opened 2018-01-10
Report all scope errors at once when defining a task.

tomprince:


djmitche:
Oops!  Scopesets accomplish this result much more generally, and that has landed since Jan 10.

PR #243, opened 2018-01-19
Bug 1431750 - add additionalProperties to schemas where missing

petemoore:
Let's wait for the CI to run on this before merging! I think I caught them all.

### DIFF
--- a/schemas/post-artifact-response.yml
+++ b/schemas/post-artifact-response.yml
@@ -3,7 +3,6 @@ title:              "Post Artifact Response"
 description: |
   Response to a request for posting an artifact.
   Note that the `storageType` property is referenced in the request as well.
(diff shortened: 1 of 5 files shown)


schemas/post-artifact-response.yml
```diff
@@ -3,7 +3,6 @@ title:              "Post Artifact Response"
 description: |
   Response to a request for posting an artifact.
   Note that the `storageType` property is referenced in the request as well.
-type:               object
```
jonasfj: again, I have mixed feelings about this change...

jonasfj:
Wow, I'm surprised how many places we forgot this...

I know it's one of the things I've been looking for... and surprised that we missed it this must @petemoore ++

PR #244, opened 2018-01-20
Update to new taskcluster-lib-api with scope expressions

imbstack:
This probably won't be the first thing I'll land, but is indicative of the type of changes we'll make!

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -547,21 +547,35 @@ api.declare({
   name:       'createTask',
   stability:  API.stability.stable,
   idempotent: true,
-  scopes:     [
-    // Legacy scope options to be removed, hence, no-longer documented
-    // [
-    //   'queue:create-task:<provisionerId>/<workerType>',
-    // ], [
-    //   'queue:define-task:<provisionerId>/<workerType>',
-    //   'queue:task-group-id:<schedulerId>/<taskGroupId>',
-    //   'queue:schedule-task:<schedulerId>/<taskGroupId>/<taskId>',
-    // ],
-    [
-      'queue:create-task:<priority>:<provisionerId>/<workerType>',
-      'queue:scheduler-id:<schedulerId>',
-    ],
-  ],
-  deferAuth:  true,
+  scopes: {AllOf: [
+    {for: 'scope', in: 'scopes', each: '<scope>'},
+    {for: 'route', in: 'routes', each: 'queue:route:<route>'},
+    {AnyOf: [
+      {AllOf: [
+        'queue:scheduler-id:<schedulerId>',
+        {AnyOf: [
+          {
+            for: 'priority',
+            in: 'priorities',
+            each: 'queue:create-task:<priority>:<provisionerId>/<workerType>',
+          },
+        ]},
+      ]},
+      {
+        if: 'legacyScopes',
+        then: {AnyOf: [
+          'queue:create-task:<provisionerId>/<workerType>',
+          {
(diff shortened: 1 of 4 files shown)


src/api.js
```diff
@@ -547,21 +547,35 @@ api.declare({
...
-    //   'queue:task-group-id:<schedulerId>/<taskGroupId>',
-    //   'queue:schedule-task:<schedulerId>/<taskGroupId>/<taskId>',
-    // ],
-    [
-      'queue:create-task:<priority>:<provisionerId>/<workerType>',
-      'queue:scheduler-id:<schedulerId>',
-    ],
-  ],
-  deferAuth:  true,
+  scopes: {AllOf: [
```
jonasfj: Consider refactoring this to be:
`AnyOf: [ {AllOf: [current_scopes]}, {AllOf: [legacy_scopes]}]`
I think that'll read easier..

src/api.js
```diff
@@ -1274,16 +1237,13 @@ api.declare({
   route:      '/poll-task-url/:provisionerId/:workerType',
   name:       'pollTaskUrls',
   stability:  API.stability.stable,
-  scopes: [
-    [
-      // Legacy scopes
+  scopes: {AnyOf: [
+    'queue:poll-task-urls:<provisionerId>/<workerType>',
```
jonasfj: for consistency in docs, I suggest prefixing variables with `task.` even if that means we have to add `task: {...}` when we do params...

jonasfj: hmm, maybe not, because there is a lot of places where this makes less sense...


maybe we should do it the other way around and get rid of `task.` ?

src/api.js
djmitche: I'm tempted to suggest supporting comments in scope expressions, just to see the look on @imbstack's face.  But seriously, I like this arrangement since it makes it clear what changed.  What if the boolean condition was called `legacyRequestWithoutPriority` or something to suggest this is not preferred?

src/api.js
djmitche: It's not been qualified in the docs so far, without causing a lot of confusion, so I think that we should omit it.

src/api.js
```diff
@@ -938,16 +922,13 @@ api.declare({
   route:      '/task/:taskId/schedule',
   name:       'scheduleTask',
   stability:  API.stability.stable,
-  scopes:     [
-    [
-      // Legacy scope
+  scopes:     {AnyOf: [
+    'queue:schedule-task:<task.schedulerId>/<task.taskGroupId>/<taskId>',
+    {AllOf: [ // Legacy scopes
```
djmitche: could this have `{if: 'allowLegacyScopes', then: ...}` to suggest that these are legacy scopes (with `allowLegacyScopes: true` in the API implementation).

src/api.js
imbstack: Accurate rendering of the look on my face: 🤕

I do like the idea of rearranging and changing the name of the boolean though.

djmitche: To clarify: I like the arrangement in the PR commits over the arrangement Jonas proposed.

src/api.js
imbstack: We can do this no-problem when we already call `req.authorize`, but to add it to functions that don't defer auth, we will need to add an otherwise unnecessary `req.authorize` call. Would you still prefer it?

djmitche: Yes, I think so

PR #251, opened 2018-02-28
Bug  1437015 - add clientId

fiennyangeln:
[Bug 1437015](https://bugzilla.mozilla.org/show_bug.cgi?id=1437015)
Queue.task response should include clientId that created task

### DIFF
--- a/schemas/create-task-request.yml
+++ b/schemas/create-task-request.yml
@@ -18,9 +18,7 @@ properties:
     description: |
       Unique identifier for a client, who creates the task
     type:           string
-    minLength:      {$const: identifier-min-length}
-    maxLength:      {$const: identifier-max-length}
-    pattern:        {$const: identifier-pattern}
+    pattern:        "^[A-Za-z0-9!@/:.+|_-]+$"
   workerType:
     title:          "Worker Type"
     description: |
--- a/schemas/create-task-request.yml
+++ b/schemas/create-task-request.yml
@@ -13,6 +13,14 @@ properties:
     minLength:      {$const: identifier-min-length}
     maxLength:      {$const: identifier-max-length}
     pattern:        {$const: identifier-pattern}
+  clientId:
+    title:          "Client Id"
+    description: |
+      Unique identifier for a client, who creates the task
+    type:           string
+    minLength:      {$const: identifier-min-length}
+    maxLength:      {$const: identifier-max-length}
+    pattern:        {$const: identifier-pattern}
   workerType:
     title:          "Worker Type"
     description: |
--- a/schemas/task.yml
+++ b/schemas/task.yml
@@ -18,9 +18,7 @@ properties:
     description: |
       Unique identifier for a client, who creates the task
     type:           string
-    minLength:      {$const: identifier-min-length}
-    maxLength:      {$const: identifier-max-length}
-    pattern:        {$const: identifier-pattern}
+    pattern:        "^[A-Za-z0-9!@/:.+|_-]+$"
   workerType:
     title:          "Worker Type"
     description: |
--- a/schemas/task.yml
+++ b/schemas/task.yml
@@ -13,6 +13,14 @@ properties:
     minLength:      {$const: identifier-min-length}
     maxLength:      {$const: identifier-max-length}
     pattern:        {$const: identifier-pattern}
+  clientId:
+    title:          "Client Id"
+    description: |
+      Unique identifier for a client, who creates the task
+    type:           string
+    minLength:      {$const: identifier-min-length}
+    maxLength:      {$const: identifier-max-length}
+    pattern:        {$const: identifier-pattern}
   workerType:
     title:          "Worker Type"
     description: |
--- a/schemas/task.yml
+++ b/schemas/task.yml
@@ -223,6 +223,7 @@ properties:
 additionalProperties: false
 required:
   - provisionerId
+  - clientId
   - workerType
   - schedulerId
   - taskGroupId
--- a/src/api.js
+++ b/src/api.js
@@ -608,8 +608,7 @@ api.declare({
 }, async function(req, res) {
   var taskId  = req.params.taskId;
   var taskDef = req.body;
-  var clientId_default = await
-    req.clientId();
+  var clientId = await req.clientId();
 
   // Patch default values and validate timestamps
   var detail = patchAndValidateTaskDef(taskId, taskDef);
@@ -664,7 +663,7 @@ api.declare({
     var task = await this.Task.create({
       taskId:             taskId,
       provisionerId:      taskDef.provisionerId,
-      clientId:           taskDef.clientId || clientId_default, 
+      clientId:           taskDef.clientId === clientId && taskDef.clientId || clientId, 
       workerType:         taskDef.workerType,
       schedulerId:        taskDef.schedulerId,
       taskGroupId:        taskDef.taskGroupId,
@@ -696,7 +695,7 @@ api.declare({
     let def   = await task.definition();
 
     // Compare the two task definitions
-    if (!_.isEqual(taskDef, def)) {
+    if (taskDef.hasOwnProperty('clientId') === def.hasOwnProperty('clientId') && !_.isEqual(taskDef, def)) {
       return res.reportError('RequestConflict', [
         'taskId {{taskId}} already used by another task.',
         'This could be the result of faulty idempotency!',
@@ -844,13 +843,13 @@ api.declare({
     taskDef.schedulerId,
     deadline
   );
-  var clientId_default = await req.clientId();
+  var clientId = await req.clientId();
   // Try to create Task entity
   try {
     var task = await this.Task.create({
       taskId:             taskId,
       provisionerId:      taskDef.provisionerId,
-      clientId:           taskDef.clientId || clientId_default, 
+      clientId:           taskDef.clientId === clientId && taskDef.clientId || clientId,
       workerType:         taskDef.workerType,
       schedulerId:        taskDef.schedulerId,
       taskGroupId:        taskDef.taskGroupId,
--- a/src/api.js
+++ b/src/api.js
@@ -610,6 +610,12 @@ api.declare({
   var taskDef = req.body;
   var clientId = await req.clientId();
 
+  if (taskDef.hasOwnProperty(clientId) && taskDef.clientId !== clientId) {
+    return res.reportError('InputError', 'If supplied, clientId must be the clientId used to make the API call', {});
+  }
+
+  taskDef.clientId = clientId;
+
   // Patch default values and validate timestamps
   var detail = patchAndValidateTaskDef(taskId, taskDef);
   if (detail) {
@@ -663,7 +669,7 @@ api.declare({
     var task = await this.Task.create({
       taskId:             taskId,
       provisionerId:      taskDef.provisionerId,
-      clientId:           taskDef.clientId === clientId && taskDef.clientId || clientId, 
+      clientId:           clientId, 
       workerType:         taskDef.workerType,
       schedulerId:        taskDef.schedulerId,
       taskGroupId:        taskDef.taskGroupId,
@@ -695,7 +701,7 @@ api.declare({
     let def   = await task.definition();
 
     // Compare the two task definitions
-    if (taskDef.hasOwnProperty('clientId') === def.hasOwnProperty('clientId') && !_.isEqual(taskDef, def)) {
+    if (!_.isEqual(taskDef, def)) {
       return res.reportError('RequestConflict', [
         'taskId {{taskId}} already used by another task.',
         'This could be the result of faulty idempotency!',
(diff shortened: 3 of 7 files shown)


src/api.js
```diff
@@ -608,6 +608,8 @@ api.declare({
 }, async function(req, res) {
   var taskId  = req.params.taskId;
   var taskDef = req.body;
+  var clientId_default = await
+    req.clientId();
```
djmitche: This whole statement should be on a single line

src/api.js
```diff
@@ -662,6 +664,7 @@ api.declare({
     var task = await this.Task.create({
       taskId:             taskId,
       provisionerId:      taskDef.provisionerId,
+      clientId:           taskDef.clientId || clientId_default,
```
djmitche: @fiennyangeln we need to check that `taskDef.clientId === clientId_default` first, otherwise a caller could "lie" about their clientId.

@petemoore this seems a bit silly.  Why not just fill the field in with `req.clientId` unconditionally, instead of allowing the user to specify it in taskDef?

src/api.js
```diff
@@ -608,6 +608,8 @@ api.declare({
 }, async function(req, res) {
   var taskId  = req.params.taskId;
   var taskDef = req.body;
+  var clientId_default = await
```
djmitche: We don't want to mix capitalization styles.  This would be spelled `clientIdDefault` instead.  But, it's not just the default, it's the only option, so perhaps just `clientId` or `callerClientId`?

src/data.js
```diff
@@ -11,6 +11,7 @@ let Task = Entity.configure({
   properties: {
     taskId:         Entity.types.SlugId,
     provisionerId:  Entity.types.String,
+    clientId:       Entity.types.String,
```
djmitche: When modifying a data format in `data.js`, it's necessary to add a new [version](https://github.com/taskcluster/azure-entities#migrations).  This table already has a few versions, so just add one more at the end.

djmitche:
Ah, in fact, those definitions differ now, so the tests are failing with

```
     Error: taskId liAR2PURSxGri2rZr31J1g already used by another task.
This could be the result of faulty idempotency!
Existing task definition was:
 ```js
{
  "provisionerId": "no-provisioner",
  "clientId": "test-client",
  "workerType": "query-test-worker",
  "schedulerId": "my-scheduler",
  "taskGroupId": "dSlITZ4yQgmvxxAi4A8fHQ",
  "dependencies": [],
  "requires": "all-completed",
  "routes": [],
  "priority": "lowest",
  "retries": 5,
  "created": "2018-02-28T20:44:23.953Z",
  "deadline": "2018-02-28T20:46:23.953Z",
  "expires": "2019-02-28T20:46:23.953Z",
  "scopes": [],
  "payload": {},
  "metadata": {
    "name": "Unit testing task",
    "description": "Task created during unit tests",
    "owner": "jonsafj@mozilla.com",
    "source": "https://github.com/taskcluster/taskcluster-queue"
  },
  "tags": {
    "purpose": "taskcluster-testing"
  },
  "extra": {}
}
```
This request tried to define:
 ```js
{
  "provisionerId": "no-provisioner",
  "workerType": "query-test-worker",
  "schedulerId": "my-scheduler",
  "taskGroupId": "dSlITZ4yQgmvxxAi4A8fHQ",
  "routes": [],
  "retries": 5,
  "created": "2018-02-28T20:44:23.953Z",
  "deadline": "2018-02-28T20:46:23.953Z",
  "scopes": [],
  "payload": {},
  "metadata": {
    "name": "Unit testing task",
    "description": "Task created during unit tests",
    "owner": "jonsafj@mozilla.com",
    "source": "https://github.com/taskcluster/taskcluster-queue"
  },
  "tags": {
    "purpose": "taskcluster-testing"
  },
  "dependencies": [],
  "requires": "all-completed",
  "priority": "lowest",
  "extra": {},
  "expires": "2019-02-28T20:46:23.953Z"
}
```
----
method:     createTask
errorCode:  RequestConflict
statusCode: 409
time:       2018-02-28T20:44:24.153Z
```

fiennyangeln:
@djmitche should i modify all the other test (the task def inside each of them), i only modify the createtas_test so far. Or should we use the default clientId instead?

src/api.js
fiennyangeln: ok will modify

src/data.js
fiennyangeln: ok will modify.

fiennyangeln:
@djmitche Hi, I have modified the code and test it for all related tests not only the createtask_test. please review.

src/data.js
```diff
@@ -222,12 +222,70 @@ let Task = Entity.configure({
...
+     * - resolved
+     * See schema for task status structure for details.
+     * Remark that `runId` always match the index in the array.
+     */
+    runs:               Entity.types.JSON,
+    /** Time at which claim to latest run expires, new Date(0) if none */
+    takenUntil:         Entity.types.Date,
+  },
+  migrate(item) {
+    item.clientId = 'default';
```
djmitche: Let's set this to `unknown`.

src/api.js
```diff
@@ -662,6 +663,7 @@ api.declare({
     var task = await this.Task.create({
       taskId:             taskId,
       provisionerId:      taskDef.provisionerId,
+      clientId:           taskDef.clientId === clientId && taskDef.clientId || clientId,
```
djmitche: I think it would be easier -- and clearer -- to just use `clientId` here.  That will *always* be the correct value.

As written, this is saying "x if x is the same as y, otherwise y", which is a long way of saying "y" :)

Also, if the user provides an incorrect clientId, this conditional will just ignore it, rather than returning an error.

schemas/task.yml
```diff
@@ -13,6 +13,14 @@ properties:
...
     maxLength:      {$const: identifier-max-length}
     pattern:        {$const: identifier-pattern}
+  clientId:
+    title:          "Client Id"
+    description: |
+      Unique identifier for a client, who creates the task
+    type:           string
+    minLength:      {$const: identifier-min-length}
+    maxLength:      {$const: identifier-max-length}
+    pattern:        {$const: identifier-pattern}
```
djmitche: ClientIds have a different schema than other identifiers: `pattern:  "^[A-Za-z0-9!@/:.+|_-]+$"`.  There's no minimum or maximum length.

I forgot to mention this in the last review, sorry.

src/api.js
```diff
@@ -608,6 +608,7 @@ api.declare({
 }, async function(req, res) {
   var taskId  = req.params.taskId;
   var taskDef = req.body;
+  var clientId = await req.clientId();
 
   // Patch default values and validate timestamps
   var detail = patchAndValidateTaskDef(taskId, taskDef);
```
djmitche: Right around here would be a good place to add a check of the supplied clientId -- if supplied, it must equal `clientId`, otherwise return `res.reportError('InputError', 'If supplied, clientId must be the clientId used to make the API call', {})`.

src/data.js
```diff
@@ -222,12 +222,70 @@ let Task = Entity.configure({
     }
     return item;
   },
+}).configure({
+  version:              5,
```
djmitche: :+1:

src/api.js
```diff
@@ -693,7 +695,7 @@ api.declare({
     let def   = await task.definition();
 
     // Compare the two task definitions
-    if (!_.isEqual(taskDef, def)) {
+    if (taskDef.hasOwnProperty('clientId') === def.hasOwnProperty('clientId') && !_.isEqual(taskDef, def)) {
```
djmitche: If the clientIds were different, I think this would just skip comparing the rest of the data.  We don't want that!

It does make sense to verify that the clientIds match here.  Perhaps the right thing to do is to set `taskDef.clientId = clientId` earlier in this function (after the check I suggested above).  Then both `taskDef` and `def` will always have a clientId, and the `_.isEqual` will be enough.

fiennyangeln:
@djmitche Hi Dustin! :) Modified. Thanks for the input :)

fiennyangeln:
@djmitche I have added the required field inside the schema. Yeah sure!! I will take a look at the tools repo.

djmitche:
I pushed an update to the descriptions :)

fiennyangeln:
@djmitche thank you. can i take other task now / should I think of way to expose the clientId in the tools repo?

djmitche:
Probably the tools repo is the best next step -- without it, this change doesn't serve much purpose :)

fiennyangeln:
ok!:) Thanks for the guides!

fiennyangeln:
@djmitche Hi i have taken a look at the tools repo. Will it be placed inside the index browser and Task & Group Inspector?

djmitche:
Just the Task & Group Inspector

schemas/create-task-request.yml
```diff
@@ -13,6 +13,15 @@ properties:
     minLength:      {$const: identifier-min-length}
     maxLength:      {$const: identifier-max-length}
     pattern:        {$const: identifier-pattern}
+  clientId:
```
jonasfj: @djmitche, why is this part of the task definition?

Could we maybe make it part of `task.metadata` and then just infer it from the caller?

I know this is a design choice, so I'm curious what the pros/cons are...

---
Other question, when issuing temporary credentials to workers in `claimWork` should we re-use the task `clientId`?

That way if I use `my-client-id` to create taskA and taskA uses `task.scopes` to do something bad the audit log will say `my-client-id` did it.

Sorry, I'm bring up the design here, but implementation is probably a detail, agreeing what we should do here is the harder part.

src/api.js
```diff
@@ -608,6 +608,13 @@ api.declare({
 }, async function(req, res) {
   var taskId  = req.params.taskId;
   var taskDef = req.body;
+  var clientId = await req.clientId();
+
+  if (taskDef.hasOwnProperty(clientId) && taskDef.clientId !== clientId) {
```
jonasfj: This should have a comment, and move it into `patchAndValidateTaskDef`

jonasfj:
@petemoore, what is the design rationale behind this?
What do you want it for? What do you aim to achieve long term? ie. is this the basis for other changes in the future, like better traceability, setting different `clientId` on the temporary credentials being created in `claimWork`?

What are the pros/cons on design choices?

----
I hate to stop the party, but I'm conservative about adding stuff without a fully thought through rationale.
Clearly, `clientId` has to do with traceability of who authorized something, but so far we rely exclusively on the audit-log to help us with this. I see us adding `clientId` to resources like tasks, artifacts, secrets, workerTypes, roles and clients, so we can see who created/modified a resource.

But before we jump in and halfway do something in direction, I would like to see a fully thought through design strategy. I'm not convinced this appropriate scope for a good first bug.

jonasfj:
Note: I'm not saying this isn't the right thing to do. I'm saying I don't want to use `clientId` as a concept without a higher final goal in mind. We don't add properties left and right, just because it might be neat.

schemas/create-task-request.yml
djmitche: Putting it in metadata's not a bad idea (and not a large change to the implementation).

I'd be OK using this as *part* of the clientId in the claim credentials, but those already include a taskId so it would really only save a step -- and in situations where task A creates task B creates task C .. it could potentially produce very long clientIds.  Probably best to avoid that.

src/api.js
fiennyangeln: @jonasfj Ok! :D modified

petemoore:
> @petemoore, what is the design rationale behind this?
> What do you want it for? What do you aim to achieve long term? ie. is this the basis for other changes in the future, like better traceability, setting different `clientId` on the temporary credentials being created in `claimWork`?
> 
> What are the pros/cons on design choices?

There is some context for this in [bug 1437015](https://bugzilla.mozilla.org/show_bug.cgi?id=1437015)

> I hate to stop the party, but I'm conservative about adding stuff without a fully thought through rationale.
> Clearly, `clientId` has to do with traceability of who authorized something, but so far we rely exclusively on the audit-log to help us with this. I see us adding `clientId` to resources like tasks, artifacts, secrets, workerTypes, roles and clients, so we can see who created/modified a resource.
> 
> But before we jump in and halfway do something in direction, I would like to see a fully thought through design strategy. I'm not convinced this appropriate scope for a good first bug.

@jonasfj @djmitche and I will meet tomorrow (6 March) to discuss this face-to-face.

jonasfj:
We had discussion about this on the team meeting this morning.
I don't think we reached consensus. I think some of us realized that if want to trace requests in the system we want a more comprehensive solution. Notably there is some fear that adding a partial solution could do more harm than good, and leave us with legacy if we some day do a comprehensive solution. Especially, considering the lack of use-cases for this feature at the moment.

@fiennyangeln, for the record there is nothing with the work you did! It's pretty amazing to jump in like that and find the right places to tweak. 

I should probably have voiced my concerns when I saw the discussion of this bug on IRC, and stopped it from becoming a good-first-bug. I hope you stick around, we'll definitely try our best to find less controversial topics :)

fiennyangeln:
Ok thank you @jonasfj !!:D

PR #252, opened 2018-03-09
Upgrade taskcluster-lib-api version

fiennyangeln:
Upgrade taskcluster-lib-api version to the newest one. It is to
enable Access-Control-Max-Age, add name as options as it 
is required.

Related to Bug 1442636

@djmitche please review:D


djmitche:
Deployed!

PR #254, opened 2018-04-03
Bug 1449848 - Escape markdown (by upgrading tc-lib-api)

djmitche:
This brings in https://github.com/taskcluster/taskcluster-lib-api/pull/91


djmitche:
They're disabled for PRs because there aren't credentials..

Issue #234, opened 2017-11-27
Intermittent failure of test "putDeadlineMessage, pollDeadlineQueue"


petemoore:
Test `"putDeadlineMessage, pollDeadlineQueue"` fails intermittently. I suspect this may be due to a race condition caused by an aggressive timeout:

https://github.com/taskcluster/taskcluster-queue/blob/30a456a6cc8c3660631d3780060a7e6040431cce/test/queueservice_test.js#L56

petemoore:
@jonasfj what do you think?

petemoore:
The assertion looks for a task with some matching properties in a list, and doesn't find a task that matches - in this case, it would be useful if the test outputted the criteria it was searching for, together with the list of tasks it got back, so the user can see what tasks were returned if any, and why they didn't match. This is especially useful for intermittent tests, where a failure might not be trivial to reproduce.

jonasfj:
Nope, that timeout is supposed to happen...

It puts a message and then waits for it to show up after the timeout...
It could be intermittent if multiple unit tests are running concurrently...

djmitche:
I think this intermittency has gotten fixed along the road to making the tests better..

djmitche:
..and of course, that fails on my very next `yarn test` run.

https://bugzilla.mozilla.org/show_bug.cgi?id=1453514 for better visibility and tracking.

PR #258, opened 2018-04-11
Remove custom InputError definition, since lib-api supports it

djmitche:
..just doing what the comment says!


djmitche:
In my testing, only the intermittent test (#234) failed.

PR #259, opened 2018-04-11
Bug 1434322 - return 404's when appropriate and always upsert upstream objects

djmitche:
Plus a bunch of refactoring..


djmitche:
I did run the tests.  It doesn't run on PRs because they don't have access to secrets.  That will get cleaned up as we come up to the new testing standards.

PR #260, opened 2018-04-11
Bug 1453514 - disable intermittent test

djmitche:
If you want to take a look at the bug and fix the intermittency, we can re-enable when that's done..


jonasfj:
Okay, we can fix that later I agree...

PR #262, opened 2018-04-12
Reduced duplication of actions in schemas, and updated titles and descriptions

petemoore:
Hi guys!

I don't actually need three reviews - but I know all three of you might care about these changes, so requesting review from all of you in case you want to comment.

Note, you might want to wait for CI to complete before reviewing, as I'll be relying on it to gain assurance I've not broken anything.

Pete

### DIFF
--- a/schemas/constants.yml
+++ b/schemas/constants.yml
@@ -136,23 +136,35 @@ dependency-relation:
 
 # Description of actions. They are defined on the provisioner
 action-description: |
-  Actions are defined on the provisioner and they are performed via a request using the `url` and `method`.
-  Depending on the `context`, the `url` will be filled with parameters. In addition, `context` is
-  used by the front-end to know where to display the action. For example, `context=worker`
-  will display the action on the worker explorer.
+  Actions provide a generic mechanism to expose additional features of a
+  provisioner, worker type, or worker to Taskcluster clients.
 
-  _Note: The request will be signed with the user's Taskcluster credentials._
+  An action is comprised of metadata describing the feature it exposes,
+  together with a webhook for triggering it.
 
-# Description of the action context
-action-context-description: |
-  Actions have a "context" that is one of provisioner, worker-type,
-  or worker, indicating which it applies to. `context` is used by the front-end to know where to display the action.
+  The Taskcluster tools site, for example, retrieves actions when displaying
+  provisioners, worker types and workers. It presents the provisioner/worker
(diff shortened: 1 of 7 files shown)


schemas/constants.yml
```diff
@@ -136,23 +136,31 @@ dependency-relation:
...
-  | provisioner | Provisioner Explorer  |
-  | worker-type | Workers Explorer      |
-  | worker      | Worker Explorer       |
+  Actions provide a generic mechanism to expose additional features of a
+  provisioner, worker type, or worker to taskcluster clients.
+
+  An action is comprised of metadata describing the feature it exposes,
+  together with a webhook for triggering it.
+
+  The taskcluster tools site retrieves actions when displaying provisioners,
```
djmitche: ", or any other system that understands [actions](https://docs.taskcluster.net/reference/platform/taskcluster-queue/docs/actions),"

schemas/worker-response.yml
```diff
@@ -100,7 +101,8 @@ properties:
             Appropriate title for any sort of Modal prompt.
         context:
           title:  "Context"
-          description: {$const: action-context-description}
+          description: |
+            The value `worker`.
```
djmitche: "Only actions with the context `worker` are included."

schemas/workertype-response.yml
```diff
@@ -65,7 +66,8 @@ properties:
             Appropriate title for any sort of Modal prompt.
         context:
           title:  "Context"
-          description: {$const: action-context-description}
+          description: |
+            The value `worker-type`.
```
djmitche: "Only actions with the context `worker-type` are included."

schemas/constants.yml
```diff
@@ -136,23 +136,31 @@ dependency-relation:
...
-  Actions have a "context" that is one of provisioner, worker-type,
-  or worker, indicating which it applies to. `context` is used by the front-end to know where to display the action.
-
-  | `context`   | Page displayed        |
-  |-------------|-----------------------|
-  | provisioner | Provisioner Explorer  |
-  | worker-type | Workers Explorer      |
-  | worker      | Worker Explorer       |
+  Actions provide a generic mechanism to expose additional features of a
+  provisioner, worker type, or worker to taskcluster clients.
```
djmitche: It might be better to just include one paragraph here, and link to https://docs.taskcluster.net/reference/platform/taskcluster-queue/docs/actions, as in `actions.yml`.

PR #264, opened 2018-04-16
Return 406 if Accept-Encoding doesn't match store Content-Encoding

jonasfj:
We only do this for new the `blob` artifacts, as they are not in wide-spread use yet.
Hence, as we adopt them people will have time to switch over.

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -377,14 +377,14 @@ let Artifact = Entity.configure({
      *   contentEncoding:   Content encoding, such as 'gzip',
      *                      NOTE: This property is not present if contentEncoding is 'identity'.
      *   contentSha256:     Hex encoded sha256 hash of the un-encoded artifact
-     *   contentLength:     Size of the un-encoded artifact
-     *   transferLength:    Size of the content-encoding encoded artifact,
+     *   contentLength:     Size of the un-encoded artifact as number of bytes
+     *   transferLength:    Size of the content-encoding encoded artifact as number of bytes,
      *                      NOTE: This property is not present if contentEncoding is 'identity'.
      *   transferSha256:    Hax encooded sha256 hash of the content-encoding encoded artifact,
      *                      NOTE: This property is not present if contentEncoding is 'identity'.
-     *   provider:          Always 's3'
-     *   region:            Always 'us-west-2'
-     *   bucket:            Always configured public or private bucket
+     *   provider:          Provider currently 's3'
+     *   region:            Region currently 'us-west-2'
+     *   bucket:            Bucket, currently configured public or private bucket
      *   key:               Path to the artifact in S3
      *   uploadId:          S3 multipart uploadId.
      *                      NOTE: This property is not present when upload is completed.
--- a/src/data.js
+++ b/src/data.js
@@ -372,21 +372,30 @@ let Artifact = Entity.configure({
      * storageType: error
      *   reason:        Formalized reason for error artifact, see JSON schema.
      *   message:       Human readable error message to return
+     *
      * storageType: blob
-     *   contentType
-     *   contentEncoding
-     *   contentSha256: SHA256 hash of the un-encoded artifact
-     *   contentLength: Number of bytes of the un-encoded artifact
-     *   transferSha256: SHA256 hash of the content-encoding encoded artifact
-     *   transferLength: Number of bytes of the content-encoding encoded artifact
-     *   partsHash: Rather than storing the potentially large list of sha256/size
-     *              pairings for each part, we just need to store enough information
-     *              to determine if the operation would result in an idempotency
-     *              error
+     *   contentEncoding:   Content encoding, such as 'gzip',
+     *                      NOTE: This property is not present if contentEncoding is 'identity'.
+     *   contentSha256:     Hex encoded sha256 hash of the un-encoded artifact
+     *   contentLength:     Size of the un-encoded artifact
+     *   transferLength:    Size of the content-encoding encoded artifact,
+     *                      NOTE: This property is not present if contentEncoding is 'identity'.
+     *   transferSha256:    Hax encooded sha256 hash of the content-encoding encoded artifact,
+     *                      NOTE: This property is not present if contentEncoding is 'identity'.
+     *   provider:          Always 's3'
+     *   region:            Always 'us-west-2'
+     *   bucket:            Always configured public or private bucket
+     *   key:               Path to the artifact in S3
+     *   uploadId:          S3 multipart uploadId.
+     *                      NOTE: This property is not present when upload is completed.
+     *   etag:              S3 ETag value.
+     *                      NOTE: This property is not present until upload is completed.
+     *   partsHash:         Hash of (sha256, size) tuples for all parts.
+     *                      We store this to enable idempotency, when a request is retried.
      */
     details:        Entity.types.JSON,
     expires:        Entity.types.Date,
-    /** 
+    /**
      * Present is a number field which represents an artifact being present and
      * the upload being completed.  The handling logic for the artifact's
      * storage type will fully define what that means for a given storage type.
--- a/test/artifact_test.js
+++ b/test/artifact_test.js
@@ -238,7 +238,7 @@ suite('Artifacts', function() {
 
     test('S3 single part complete flow, content-encoding: gzip', async () => {
       const taskId = slugid.v4();
-      const data = crypto.randomBytes(/*12 * 1024 * 1024 +*/ 21);
+      const data = crypto.randomBytes(12 * 1024 * 1024 + 21);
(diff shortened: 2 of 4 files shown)


src/data.js
```diff
@@ -372,21 +372,30 @@ let Artifact = Entity.configure({
...
-     *              error
+     *   contentEncoding:   Content encoding, such as 'gzip',
+     *                      NOTE: This property is not present if contentEncoding is 'identity'.
+     *   contentSha256:     Hex encoded sha256 hash of the un-encoded artifact
+     *   contentLength:     Size of the un-encoded artifact
+     *   transferLength:    Size of the content-encoding encoded artifact,
+     *                      NOTE: This property is not present if contentEncoding is 'identity'.
+     *   transferSha256:    Hax encooded sha256 hash of the content-encoding encoded artifact,
+     *                      NOTE: This property is not present if contentEncoding is 'identity'.
+     *   provider:          Always 's3'
```
jhford: Knowing this is very likely to change at some point in this year, let's say "Currently always 's3' but might change later"

src/data.js
```diff
@@ -372,21 +372,30 @@ let Artifact = Entity.configure({
...
+     *   region:            Always 'us-west-2'
```
jhford: Knowing this is very likely to change at some point in this year, let's say "Currently always 'us-west-2' but might change later"

src/data.js
```diff
@@ -372,21 +372,30 @@ let Artifact = Entity.configure({
...
+     *   bucket:            Always configured public or private bucket
```
jhford: Maybe note that while we're calling it `bucket`, that it's whatever the equivalent concept is for the backing service.

src/data.js
```diff
@@ -372,21 +372,30 @@ let Artifact = Entity.configure({
...
+     *   key:               Path to the artifact in S3
```
jhford: Maybe note that while we're calling it key, that it's whatever the equivalent concept is for the backing service.

src/data.js
```diff
@@ -372,21 +372,30 @@ let Artifact = Entity.configure({
...
+     *   uploadId:          S3 multipart uploadId.
+     *                      NOTE: This property is not present when upload is completed.
```
jhford: It should also only ever be present if the upload was an S3 multipart upload

src/data.js
```diff
@@ -372,21 +372,30 @@ let Artifact = Entity.configure({
...
+     *   etag:              S3 ETag value.
```
jhford: ETags aren't S3 specific, let's just say that we're storing the Etag (if available) of the backing service's view of the object.

test/artifact_test.js
```diff
@@ -227,18 +228,79 @@ suite('Artifacts', function() {
...
         `https://test-bucket-for-any-garbage.s3-us-west-2.amazonaws.com/${taskId}/0/public/singlepart.dat`;
       assume(artifact.headers).has.property('location', expectedUrl);
 
       await verifyDownload(artifact.headers.location, bigfilehash, bigfilesize);
 
     });
 
+    test('S3 single part complete flow, content-encoding: gzip', async () => {
+      const taskId = slugid.v4();
+      const data = crypto.randomBytes(/*12 * 1024 * 1024 +*/ 21);
```
jhford: I'm curious why the `12*1024*1024` in comments and why 21bytes.

test/artifact_test.js
```diff
@@ -227,18 +228,79 @@ suite('Artifacts', function() {
...
+        contentSha256: crypto.createHash('sha256').update(data).digest('hex'),
+        transferLength: gzipped.length,
+        transferSha256: crypto.createHash('sha256').update(gzipped).digest('hex'),
+        parts: [{
+          size: gzipped.length,
+          sha256: crypto.createHash('sha256').update(gzipped).digest('hex'),
+        }],
+      });
+
+      debug('### Put first part of artifact');
```
jhford: do you mean the only part :)

test/artifact_test.js
```diff
@@ -227,18 +228,79 @@ suite('Artifacts', function() {
...
+      const {method, url, headers} = requests[0];
```
jhford: Please match the way that all the other blob-type unit tests run uploads instead of reimplementing it.

```javascript
      let uploadOutcome = await client.runUpload(response.requests, uploadInfo);
```

test/artifact_test.js
```diff
@@ -227,18 +228,79 @@ suite('Artifacts', function() {
...
+      } = await helper.queue.createArtifact(taskId, 0, 'public/singlepart.dat', {
+        storageType: 'blob',
+        expires: taskcluster.fromNowJSON('1 day'),
+        contentType: 'binary/octet-stream',
+        contentEncoding: 'gzip',
+        contentLength: data.length,
+        contentSha256: crypto.createHash('sha256').update(data).digest('hex'),
+        transferLength: gzipped.length,
+        transferSha256: crypto.createHash('sha256').update(gzipped).digest('hex'),
+        parts: [{
```
jhford: This is testing single part uploads according to the test name and all the artifact names used, but you're doing a multipart upload with a single part.  Please delete the `parts` list from the request to do a true single part upload.

test/artifact_test.js
```diff
@@ -227,18 +228,79 @@ suite('Artifacts', function() {
...
+      await helper.queue.completeArtifact(taskId, 0, 'public/singlepart.dat', {
+        etags: [etag],
+      });
+
+      const artifactUrl = helper.queue.buildUrl(
+        helper.queue.getArtifact,
+        taskId, 0, 'public/singlepart.dat',
+      );
+      debug('### Fetching artifact from: %s', artifactUrl);
+      const res2 = await request.get(artifactUrl).responseType('blob');
```
jhford: There's already code for doing this:
```javascript
      await verifyDownload(artifact.headers.location, bigfilehash, bigfilesize);
```

Please use this code as it's what we do for all the other tests and handles many more tests than are duplicated here.

test/artifact_test.js
```diff
@@ -227,18 +228,79 @@ suite('Artifacts', function() {
...
+      debug('Downloaded artifact, statusCode: %s', res2.status);
+      const res2Hash = crypto.createHash('sha256').update(res2.body).digest('hex');
+      assert(res2Hash === crypto.createHash('sha256').update(data).digest('hex'));
+      assert(res2Hash === res2.headers['x-amz-meta-content-sha256']);
+      debug('Response headers: %j', res2.headers);
+
+      debug('### Downloading artifact with incorrect "Accept-Encoding"');
+      let e;
+      await request.get(artifactUrl).set('Accept-Encoding', 'identity').catch(err => e = err);
```
jhford: Let's also test for no `Accept-Encoding` value set, since that's by far the more likely scenario in reality.

src/data.js
```diff
@@ -372,21 +372,30 @@ let Artifact = Entity.configure({
...
-     *   contentSha256: SHA256 hash of the un-encoded artifact
-     *   contentLength: Number of bytes of the un-encoded artifact
-     *   transferSha256: SHA256 hash of the content-encoding encoded artifact
-     *   transferLength: Number of bytes of the content-encoding encoded artifact
-     *   partsHash: Rather than storing the potentially large list of sha256/size
-     *              pairings for each part, we just need to store enough information
-     *              to determine if the operation would result in an idempotency
-     *              error
+     *   contentEncoding:   Content encoding, such as 'gzip',
+     *                      NOTE: This property is not present if contentEncoding is 'identity'.
```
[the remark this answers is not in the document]
jonasfj: If I look at the code yes... maybe we have default in the schema and I missed that.. but then we have unnecessary special cases in the code.

test/artifact_test.js
jonasfj: oh, left over from debugging...

test/artifact_test.js
jonasfj: hmm, interesting... :)

I didn't notice this. let's talk about it tomorrow.

test/artifact_test.js
jonasfj: tried adding it that won't work because of content-encoding.

test/artifact_test.js
jonasfj: `identity` is the same thing as setting no `accept-encoding`.

jonasfj:
@jhford, I'm not sure how well, we want to test this... We basically have to do everything manually with `http` if you want perfect tests.

IMO this exercises all aspect of the code added. If we want better tests we'll need to write a lot of low-level, and that won't be nice to maintain.

test/artifact_test.js
jhford: but it should also be one extra line to account for us doing something incorrectly and handling the two different options differently.

djmitche:
highly bitrotted..

PR #265, opened 2018-04-16
Signed url artifact test (#233 with a couple more tests)

jhford:
This is a new set of patches for #233, so for the majority of the code, the R+ from there should carry over.  The really important part here is the signed URL test.  When I landed #233, @petemoore let me know of some fallout with signed URL processing.  I'm not sure what went wrong, so I wrote these tests.  There doesn't seem to be any trouble using signed URLs, so I'm unsure what's going on here.

What I did in this PR is add a test to ensure that generating a signed URL with the blob artifact type works, as well as finding and fixing a small bug with how single part blob artifacts are handled.  This PR also adds a unit test for signed URLs for blob storage types.

I've also added to the legacy `s3` storage type unit tests to test signed URLs.  It looks like we're missing unit tests for private artifacts, is that right @jonasfj ?


src/artifacts.js
```diff
@@ -708,9 +708,10 @@ api.declare({
         tags: {taskId, runId, name},
       });
     } else {
-      let url = await this.s3Controller.generateGetUrl({
+      let url = await this.s3Controller.generateUrl({
```
jonasfj: There is no method: `generateUrl`, or if there is I can't find the definition.

IMO, it be better to use `aws4` for URL signing here, instead of routing it via. `remotely-signed-s3`, because I have to cross reference to know what this does at HTTP level.

src/artifacts.js
```diff
@@ -727,7 +728,15 @@ api.declare({
...
-        throw new Error('S3 object does not have correct Content-Sha256 metadata');
+        let msg = [
+          'Expected S3 object to have SHA256 of ',
+          artifact.details.contentSha256,
+          ' but found it to have ',
+          headRes.headers['x-amz-meta-content-sha256'] || 'no value',
+          '',
+        ].join('"');
+        console.dir(headRes);
+        throw new Error(msg);
```
jonasfj: Maybe remove `console.dir` would be nice..

The right thing is probably to report internal error to sentry if this happens, as it should be possible. But we should validate that we get a  `200` response before it's an internal error. I don't mind you not doing that, because this PR doesn't touch that :)

test/artifact_test.js
```diff
@@ -235,6 +235,81 @@ suite('Artifacts', function() {
...
+        workerGroup:    'my-worker-group',
+        workerId:       'my-worker',
+      });
+
+      let uploadInfo = await client.prepareUpload({
+        filename: bigfilename,
+        forceSP: true,
+      });
+
+      let name = 'signed-url.dat';
```
jonasfj: You'll want to make a name that uses `+`  and whitespace ` ` or whatever it was charaters that require encoding that was affected by previous commits in this PR.

Probably add a bunch of special characters :)

PR #270, opened 2018-04-17
Hopefully last round of cleanup

petemoore:
This should be everything now. In total it reduces the schemas line count by 45 lines by removing duplication, which also removes the duplicated go types.

* I put common data structures into new files (`task-run.yml`, `task-credentials.yml`, `task-metadata.yml`) and `$ref` them
* Spelling correction: meta-data -> metadata
* Added missing titles that improve generated go code
* I made (`url`, `method`, `headers`) required properties, and disallowed additional properties, in Blob Artifact Response HTTP requests
* All tests pass 🕺


jonasfj:
bstack says ship it :)

PR #279, opened 2018-06-20
Bug 1462758 - upgrade queue to use newest libraries, rootUrl

djmitche:
I'm not totally happy with this, mostly because the "real" tests remain a bit intermittent.  I'm tempted to do some refactoring to fix that, but this is already a big change, and I don't think it makes sense to add risk (by refactoring) just to make tests better.  At least, not right now.

@jonasfj could you do a review as a parting gift? :)


djmitche:
This will depend on https://github.com/taskcluster/taskcluster-lib-testing/pull/30 and, I think, https://github.com/taskcluster/taskcluster-lib-testing/pull/29

djmitche:
The tc-lib-testing changes are now landed.

djmitche:
GREEEEEEEN!

src/workerinfo.js
```diff
@@ -89,7 +89,7 @@ class WorkerInfo {
           });
         }
 
-        createEntry(this.Provisioner, {
+        return createEntry(this.Provisioner, {
```
imbstack: This makes the tests much nicer!

schemas/v1/worker-response.yml
```diff
@@ -42,7 +42,7 @@ properties:
       List of 20 most recent tasks claimed by the worker.
     type:         array
     items:
-      $ref:       "http://schemas.taskcluster.net/queue/v1/task-run.json#"
+      $ref:       "task-run.json#"
```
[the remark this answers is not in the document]
djmitche: This makes a bit more sense now (even if the implementation is a little weird) -- in the API declaration, we refer to the files (implementation referring to implementation).  Everywhere else, we refer to the schema (schemas referring to schemas).  But yeah, still a little odd.

src/workerinfo.js
djmitche: And fixes a pretty major bug :)

PR #281, opened 2018-07-13
bug 1453310 -- Handle MAX_MODIFY_ATTEMPTS exhausted in claimWork

jhford:
Basically, check the error code and rethrow if it isn't MAX_MODIFY_ATTEMPTS, which should keep sentry monitoring in place.  Otherwise, return a descriptive 500 for a client retry.

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -1872,41 +1872,53 @@ api.declare({
     workerId:       run.workerId,
   });
 
-  await task.modify((task) => {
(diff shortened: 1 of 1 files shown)


jhford:
(I cannot run tests locally right now, so the r? is dependent on CI tests passing)

jhford:
closing because we'll reevaluate this in the postgres queue

PR #284, opened 2018-07-18
Bug 1476441 - fix relative docs links

djmitche:


djmitche:
Looks like the test failures here were because we output too much debug info, causing docker to lock up (https://github.com/containerd/console/pull/27)

PR #285, opened 2018-07-20
Consolidate schemas with $ref

djmitche:
This helps to clean up some documentation as it appears in the docs
service, and removes redundant, sometimes contradictory statements about
similar fields.

I started with createTask, because I had noticed the now-incorrect "Task submitter required scopes" in the docs.  But I branched out a little into some other API methods' schemas.


djmitche:
(landed, and no huge increase in 400's or 500's)

PR #287, opened 2018-08-08
Allow logging azure timings

imbstack:
As it says on the tin.

~I'll try this out in prod tonight when there's low traffic.~

Nevermind, this causes the tests to fail it seems? Maybe it was caused by the azure-entities bump. I'll look into it tomorrow!

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -12,10 +12,38 @@ tasks:
       github:
         env: true
         events:
-          - push
           - pull_request.opened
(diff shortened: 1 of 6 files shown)


imbstack:
Actually... I'm 99% sure this is the push and PR builds fighting each other. Filed [bug 1481942](https://bugzilla.mozilla.org/show_bug.cgi?id=1481942) for this. Please feel free to review now!

PR #291, opened 2018-09-27
Update lib-api to report ip addrs for auth

imbstack:


.taskcluster.yml
```diff
@@ -1,7 +1,7 @@
 version: 0
 allowPullRequests: public
 metadata:
-  name: "TaskCluster Queue Tests"
+  name: "Task:luster Queue Tests"
```
djmitche: I think that's `Task:lustre`?

imbstack: Wow. Did I really fix this by making it worse? Nice work bstack.

imbstack:
Tests pass on second run.

PR #292, opened 2018-10-05
Document why queue.rerunTask is deprecated

petemoore:
I believe this is the reason this method is deprecated.

However, I'm also aware that reruns might be useful for failures when `taskId`s are locked down by chain of trust. Creating a new task introduces the possibility that the task definition was subtly altered, whereas there is no way to alter a task definition when calling `queue.rerunTask`.

I suspect the motivation for deprecating this method was to avoid the problem that external systems may be monitoring a task resolution, and have an assumption that a failed task can't later resolve as successful, or vice-versa.

Since this is a particularly delicate concern, I'm grateful for feedback/head-nod/head-shake from any/all of @escapewindow @djmitche @imbstack.

Another common use case can be that a task depends on an external system, which intermittently fails. For example, a task could clone from hg/git, and the hg/git service may be temporarily unavailable. Rerunning a task is a useful way to gloss over that it failed the first time due to no fault of its own. For example, in the taskcluster-github integration, being able to rerun a task probably means that the annoying red cross denoting CI failure can be promoted to a nice green tick denoting success. I'm not sure if this is possible without task reruns, perhaps it forces the user to create a new commit in order to trigger the generation of a new task.

Another place this may have impact is in the "resolution" of a task group. Although you can never be sure when no more tasks will be associated with a given task group ID (although of course we could change this if we added an API endpoint to lock a task group ID to accept no new tasks), you can query the status of a task group and get back a resolution (I believe). Allowing reruns is one thing that makes the result of that query less deterministic.

So calling in the help of the experts! Thanks!

P.S. If we agree on this rationale, we should perhaps add some of the above to the docs too, so that it doesn't get lost when this PR closes.


escapewindow:
This is actually under debate.

`retrigger`s are copies of the task under a new `taskId` and a new `taskGroupId`. They can still be CoT-verifiable, if they've been created through a CoT-verifiable retrigger action. Retriggers are preferred for tests because we sometimes create multiple runs of the same task to weed out intermittent failures.

For Firefox release graphs, we prefer `rerun`s, which increment the `runId` of the current task. There are some dangers here, largely around rerunning a successfully completed task, or the fact that reruns of a task will start the task, even if its dependencies haven't completed yet. (The latter is a bug that is also a feature; we use this feature when we intentionally want to start a task before its dependencies complete.)

In releases, we need the original `taskId` status to be definitive. We monitor releases by monitoring the `taskGroupId`. Retriggers would create a separate `taskGroupId` and complicate this. Even if they created the new task inside the same `taskGroupId`, we'd have two separate `taskId`s for the same task label, and seeing 10 failed tasks in the `taskGroupId` wouldn't necessarily mean we had 10 actionable tasks; we'd also have to check they didn't exist in a separate, newer `taskId`. Also, the way we trigger the next `phase` of the release (build -> promote -> push -> ship) is to point at the previous action task's `label-to-taskid.json` to find the previous tasks to depend on. These `taskId`s are the *original* `taskId`s, so any downstream graph would be depending on failed tasks even if we had retriggered them to success previously.

Because reruns are so important to our current Firefox release process, I would argue that we either shouldn't deprecate this api method, or we should replace it with methods that replace its functionality but fix its issues. For example, `requeueBrokenTask` (or something) could mark a task as `unscheduled`, but only if its status is `exception` or `failed`. Marking it as `unscheduled` would (I think) make it still block on unresolved dependencies. `forceRerunTask` could mark any task as `pending`, regardless of task state or dependency state. A split like the above, could allow for more granular permissioning, and help avoid footguns.

djmitche:
Maybe we could record all of this rationale in the documentation (before we forget who Jonas was), without actually *deprecating* the method.  Basically, the method is supported but here's why you may want to think twice about using it.  I bet that's pretty close to what this PR looks like right now.

I like Aki's ideas, too, but that should probably be an RFC so we can trace through the consequences before committing (and doesn't block landing a doc patch like this one).

escapewindow:
I filed https://github.com/taskcluster/taskcluster-rfcs/issues/129.

petemoore:
Thanks guys. I'll update the comments to explain the reasoning, as per Jonas's arguments.

PR #294, opened 2018-10-19
Bug 1499701 - upgrade azure to behave better on SAS errors

djmitche:


djmitche:
Queue's logging is pretty crazy fast, but it the flashing on my screen looks the same before and after and no `/status=5/`, so I think this went OK.

jhford:
A really useful thing for watching the queue is to use `-d web.11` which will only show the logs of that one dyno, web.11.  Usually, a single dyno is fine for following logs on.  Since all of the 25 web dynos are the same, they have an equal chance of showing errors.

PR #295, opened 2018-10-19
Bug 1488796 - upgrade to tc-lib-pulse for sending pulse messages

djmitche:


src/main.js
```diff
@@ -25,6 +25,7 @@ let docs                = require('taskcluster-lib-docs');
 let App                 = require('taskcluster-lib-app');
 let remoteS3            = require('remotely-signed-s3');
 let {sasCredentials}    = require('taskcluster-lib-azure');
+let {Client, pulseCredentials} = require('taskcluster-lib-pulse');
```
jhford: The name `Client` is not the most specific name here.  Can this constructor be renamed to something like`PulseClient`?

src/main.js
```diff
@@ -65,15 +66,26 @@ let load = loader({
...
     }),
   },
+
+  pulseClient: {
+    requires: ['cfg', 'monitor'],
+    setup: ({cfg, monitor}) => {
+      return new Client({
+        namespace: cfg.pulse.namespace,
+        monitor,
+        credentials: pulseCredentials(cfg.pulse),
```
jhford: Not relevant to this system, but in lib-pulse, it might be a good idea to allow passing in objects in the shape of cfg.pulse and handle the mapping to a connection string internally.

src/main.js
djmitche: It can be referenced as `pulse.Client` if `const pulse = require('taskcluster-lib-pulse')`.

src/main.js
djmitche: That's what this accomplishes.  There are a few ways to provide credentials, and this is only one of them, so the *Credentials methods map them all to the same format (a callable).  This made more sense when fetching dynamic credentials from tc-pulse, but I think the model is natural enough to keep even without that feature.

djmitche:
Yes, backout is all it needs.

djmitche:
(updated to change the require format)

src/main.js
djmitche: (It's based on the model used in taskcluster-lib-azure)

djmitche:
(rebased to fix conflicts in `yarn.lock`)

djmitche:
great success!

PR #296, opened 2018-10-22
Bug 1501007 - terminate the queueservice so expirequeues exits

djmitche:


src/main.js
```diff
@@ -590,6 +590,7 @@ let load = loader({
 
       monitor.count('expire-queues.done');
       monitor.stopResourceMonitoring();
+      queueservice.terminate();
```
owlishDeveloper: `queueService`?

djmitche: It's a component that interfaces with Azure queues.
https://github.com/taskcluster/taskcluster-queue/blob/master/src/queueservice.js#L72

It has a cache, and a `setInterval` to clear that cache every 25 hours.  That interval timer causes node to continue running forever.  Calling `terminate` cancels the interval.

owlishDeveloper: I meant: isn't that a typo? Need capital `S`

djmitche: Oh, haha, I misunderstood the question.  Yes, precisely.  I apparently failed to copy this change out of the pod where I was experimenting, too.

owlishDeveloper: (thank you for the explanation tho! I didn't realize that)

imbstack:
Tests are still failing but these look more like flakes?

owlishDeveloper:
I wonder if the second failure is actually related to this change?

djmitche:
They look flaky, yeah.

PR #298, opened 2018-11-09
Bug 1486071 - simplify description of intermittent-task

djmitche:
..and replace some vague and confusing admonitions with clearer
language.

### DIFF
--- a/docs/worker-interaction.md
+++ b/docs/worker-interaction.md
@@ -146,24 +146,24 @@ it _completed_ by accident. Note. artifacts can't be added after 20 minutes or
 so.
 
 ## Intermittent Tasks
-To deal with intermittent tasks the worker can report _exception_ with the
-reason `intermittent-task`. This is only supposed to be used by tasks that
-explicitly communicate that they failed in an intermittent manner.
-Generally, this should be **strongly discouraged**, but unfortunately necessary
-in a few cases.
 
-Reporting a task _exception_ with reason `intermittent-task` will retry the task
-if retries haven't been exhausted. It is strongly encouraged that workers retry
-the task/run it already holds, rather than resolving the task and have the queue
-retry the task.
+Reporting a task _exception_ with reason `intermittent-task` will retry the
+task, using the next sequential run number, if retries haven't been exhausted.
+This is intended for use by tasks that explicitly communicate that they have
+failed in an intermittent manner and will likely succeed on a retry. It should
+be opt-in on a task-by-task basis. The next run of the task returns to the
+queue may run on the same or a different worker.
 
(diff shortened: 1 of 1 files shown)


docs/worker-interaction.md
```diff
@@ -146,22 +146,17 @@ it _completed_ by accident. Note. artifacts can't be added after 20 minutes or
...
-If using this feature, please ensure that task are explicitly asking to be
-retried, due to intermittence, and consider if perhaps the intermittence is due
-to a broken worker (dirty worker state or bad apple) and consider formatting
-the worker, reporting `worker-shutdown` instead.
 
+Reporting a task _exception_ with reason `intermittent-task` will retry the
+task if retries haven't been exhausted.
+
+This is intended for use by tasks that explicitly communicate that they failed
+in an intermittent manner and will likely succeed on a retry. It should be
```
jhford: This should distinguish between retrying on the same task/run and retrying on a completely new run on another machine.  An important point in the old wording was that there was a recommendation to favour retrying within the task run over marking as intermittent and retrying.

Maybe there should also be something mentioning that it's possible that the job will end up being run on the same worker, which might result in a repeat of the failure.

docs/worker-interaction.md
```diff
@@ -146,22 +146,17 @@ it _completed_ by accident. Note. artifacts can't be added after 20 minutes or
...
-the task/run it already holds, rather than resolving the task and have the queue
-retry the task.
-
-If using this feature, please ensure that task are explicitly asking to be
-retried, due to intermittence, and consider if perhaps the intermittence is due
-to a broken worker (dirty worker state or bad apple) and consider formatting
-the worker, reporting `worker-shutdown` instead.
 
+Reporting a task _exception_ with reason `intermittent-task` will retry the
+task if retries haven't been exhausted.
```
jhford: Can this clarify that it will use one of the runs allocated to the task?  Maybe also what happens if the final allocated run ends in intermittent-task.

docs/worker-interaction.md
```diff
@@ -146,22 +146,17 @@ it _completed_ by accident. Note. artifacts can't be added after 20 minutes or
...
-if retries haven't been exhausted. It is strongly encouraged that workers retry
-the task/run it already holds, rather than resolving the task and have the queue
-retry the task.
-
-If using this feature, please ensure that task are explicitly asking to be
-retried, due to intermittence, and consider if perhaps the intermittence is due
-to a broken worker (dirty worker state or bad apple) and consider formatting
-the worker, reporting `worker-shutdown` instead.
 
+Reporting a task _exception_ with reason `intermittent-task` will retry the
```
jhford: This should continue to discourage using intermittent-task where possible.

PR #299, opened 2018-11-20
Bug 1508395 - upgrade tc-lib-validate

djmitche:
This should only affect writeDocs


djmitche:
This test run succeeded on the second try..

PR #310, opened 2019-01-02
More monorepo fixes

imbstack:
This gets all of our tests to pass! Just needed to get xml2js updated.


imbstack:
Oh, I was bitten by aws/aws-sdk-js#1008. I'll go back to the old way but still update the lib.

